PR #1, opened 2019-11-27
Add Optional content unwrapping

mpiechocki:
Hey!
First of all, great job with this library! :) I added it to my side-project but quickly encountered the problem with optional content, so I tried to contribute a little :)

I think the test case is pretty self-explanatory.

Cheers!


nalexn:
@mpiechocki 

Nice catch on this one, and great job!

PR #8, opened 2020-02-04
Add Podfile

YuantongL:
Support CocoaPods

### DIFF
--- a/ViewInspector.podspec
+++ b/ViewInspector.podspec
@@ -1,0 +1,25 @@
+
+Pod::Spec.new do |s|
+
+  s.name = "ViewInspector"
+  s.version = "0.3.5"
+  s.summary = "ViewInspector is a library for unit testing SwiftUI views.
+               It allows for traversing a view hierarchy at runtime providing direct access to the underlying View structs."
+  s.homepage = "https://github.com/nalexn/ViewInspector"
+  s.license = { :type => "MIT", :file => "LICENSE" }
+  s.author = { "Alexey" => "https://github.com/nalexn" }
+
+  s.ios.deployment_target = '13.0'
+  s.osx.deployment_target = '10.15'
+  s.tvos.deployment_target = '13.0'
+  s.watchos.deployment_target = '6.0'
+  s.framework = 'XCTest'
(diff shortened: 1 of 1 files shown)


nalexn:
Hey @YuantongL 

Thank you for contributing. I was thinking if adding the support for CocoaPods and Carthage is actually needed.

My point is that the native tool (which is Swift Package Manager) is always preferable over any third-party solution, and given the fact that SPM is mature enough to be a stable and flexible dependency manager, CocoaPods and Carthage can be considered obsolete.

SwiftUI is a huge step forward in the iOS development, literally a revolution, and I'm not convinced we need to drag the past to the bright future.

I understand that having to mix multiple dependency managers in one project is inconvenient, but a better way would be to convince others to support SPM rather than stay with legacy solutions.

Let me know your thoughts, maybe I'm missing something.

YuantongL:
Hi @nalexn, thanks for writing this library, I really like to make use of this, our codebase uses CocoaPods only, I wouldn't consider CocoaPods legacy, CocoaPods is just another way of doing it and I'm sure there are more projects based on CocoaPods.

Also this is a single file change that can bridge your library to more audience, which I think is worth it. But it's up to you.

nalexn:
Hey @YuantongL 

Could you add my email (alexey ~at~ naumov.tech) on cocoapods for the ViewInspector. I can see the version on CocoaPods is very old, wanted to push an updated one.

YuantongL:
> Hey @YuantongL
> 
> Could you add my email (alexey >~at~< naumov.tech) on cocoapods for the ViewInspector. I can see the version on CocoaPods is very old, wanted to push an updated one.

OK, I no longer have access to my registered email, I'm filing a claim to Cocoapod. Will update you as soon as I heard back from them.

nalexn:
@YuantongL did you hear back from the Cocoapods team?

YuantongL:
> @YuantongL did you hear back from the Cocoapods team?

Not yet, I'm filing another claim now.

amorde:
Just settled this one now - thanks for your patience.

@nalexn you are now the sole owner of this Pod

PR #10, opened 2020-02-05
Added a failing test that adds multiple accessibility modifiers

nrivard:
This creates a test that adds an accessibility value and label and queries them both. It is currently failing but should be used to assert that a fix for accessibility modifiers is working properly.


nalexn:
Oops, merged before resolving. I'll do this now anyways

PR #25, opened 2020-04-15
Added support to inspect SwiftUI.Image with CGImage initializer

### LINKED ISSUE #24: Access SwiftUI.Image the CGImage based attribute
Currently we have both `.uiImage()` and `.nsImage()`. However, due to some performance or bugs of SwiftUI, we have to use the CGImage based to provide the `SwiftUI.Image`. I want to write test case to ensure this behavior.

Bugs: https://github.com/SDWebImage/SDWebImageSwiftUI/issues/101

See the API: https://developer.apple.com/documentation/swiftui/image/3364311-init

So, please make [this API](https://github.com/nalexn/ViewInspector/blob/0.3.10/Sources/ViewInspector/SwiftUI/Image.swift#L46) public, or provide the correspond API like `.cgImage()`

dreampiggy:
This close #24

### DIFF
--- a/Sources/ViewInspector/SwiftUI/Image.swift
+++ b/Sources/ViewInspector/SwiftUI/Image.swift
@@ -52,6 +52,21 @@ public extension InspectableView where View == ViewType.Image {
         }
     }
     
+    func orientation() throws -> Image.Orientation {
+        let orientation = try Inspector.attribute(path: "provider|base|orientation", value: unwrap(view: content.view)) as? Image.Orientation
+        return orientation ?? .up
+    }
+    
+    func scale() throws -> CGFloat {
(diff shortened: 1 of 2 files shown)


dreampiggy:
OK. Please wait me to add unit tests :)

dreampiggy:
Seems your project setting disable the force downcast ? So how to cast the `AnyObject` into `CGImage` or any other CoreFoundation Object ? Because they are pointer, not Swift object internlly.

image

image


Maybe I need to use `unsafeBitCast(to:)` ?

dreampiggy:
Another intersting limit :)

image

Any better idea for coding style like this

dreampiggy:
@nalexn  Test Cases added, see `testExtractionCGImage`

PR #28, opened 2020-05-22
This should probably use vStack()

twostraws:
If this test is supposed to pass (and hopefully it is, because it's the first example), I would imagine it needs `vStack()` here rather than `hStack()`.


nalexn:
Nice find, Paul! Thank you!

PR #31, opened 2020-06-18
Change platform to v11 for iOS to be able to use `ViewInspector` in apps with optional SwiftUI views

gunterhager:
We have a use case where our app has a build target below iOS 13. We use SwiftUI for views that show only on devices running at iOS 13 and above. 

I think being able to unit test views during a period of transition between UIKit and SwiftUI based apps is quite essential. Therefore it is useful to not require iOS 13 for `ViewInspector`.

I have marked all iOS 13 only features with `@available` to be able to use the package with projects that use SwiftUI only optionally.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -6,7 +6,7 @@ import PackageDescription
 let package = Package(
     name: "ViewInspector",
     platforms: [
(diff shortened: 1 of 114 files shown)


nalexn:
@gunterhager 

Thank you! Merged! New version 0.4.0

PR #35, opened 2020-07-24
fontWeight inspection support

vruzeda:
Add support for inspecting `fontWeight` in `Text` views.


vruzeda:
I had a better, more flexible idea on how to implement this.

nalexn:
I was going to tweak your variant over the weekend and merge. The main concern is that I'd ideally come up with more unified API for accessing Text modifiers, as there are many. My concern with the initial implementation was that we'd have to always deal with an array of optional values, while 99% of the time you'll have a single Text with a set of modifiers. I need to put more thoughts into this. Let me know what you have in mind

PR #36, opened 2020-07-24
attributedString inspection support

vruzeda:
Add support for a "pseudo-inspector" that converts `Text` views in `NSAttributedString` objects, making it easier to test their styles.


nalexn:
What a brilliant idea to use NSAttributedString! I'll review and merge over the weekend

nalexn:
Decided to upgrade your solution for convenience - introduced a custom container instead of the NSAttributedString. Thank you for contributing!

Issue #41, opened 2020-08-30
Verifying Text attributes based on Range<String.Index>


Tunous:
Given the following example:

```swift
let text = Text("bold").bold() + Text(" ") + Text("italic").italic()
```

I would like to verify that whole "italic" word has correct attribute applied without the need to manually calculate correct range (`5..<11` I believe). My first idea was to use `range(of:)` method on string to find the part that interests me like this:

```swift
let inspectableText = try text.inspect().text()
let string = try inspectableText.string()
let attributes = try inspectableText.attributes()
let italicTextRange = string!.range(of: "italic")!
XCTAssertTrue(try attributes[italicTextRange].isItalic()) // ERROR: Cannot convert value of type 'Range<String.Index>' to expected argument type 'Range<Int>'
```

Unfortunately that is currently not possible due to marked error.

Do you think it would be desirable to extend `TextAttributes` struct to allow such checks? I'll probably submit a proof of concept pull request if you agree. (Though I'm not yet sure if that's even possible 😄, EDIT: first attempt shows that it is possible)

Alternatively perhaps you know of another way to avoid counting letters?

Tunous:
Closing since the pull request implementing it has been merged.

PR #42, opened 2020-08-31
Support access to Text attributes based on Range<String.Index>

Tunous:
This is an implementation of idea described in #41, which allows to access `Text` attributes by using ranges created from strings.

I've implemented it as a subscript which accepts types conforming to `RangeExpression` protocol. That way it's also possible to use closed, partial or other ranges. Similar thing could be done for subscript which accepts regular `Range<Int>` but that's out of scope of this pull request.


nalexn:
Nice, @Tunous  👍👍👍

PR #63, opened 2020-12-14
Note import statements for Inspection

joemasilotti:
`import Combine` tripped me up for a little while, so I'm assuming others might benefit from being more explicit here.


nalexn:
Thank you @joemasilotti

PR #76, opened 2021-02-11
Updated the flexFrame() inspector that it also works when only value was set at the frame

renep:
When in a view only one frame parameter was set, the `flexFrame()` inspector did throw an exception. I have updated the code, so that not exception is thrown and for all the missing values `CGFloat.nan` is used. This was the simplest approach that does not brake and compatibility.


nalexn:
Awesome work on all the PRs, @renep !

PR #82, opened 2021-02-18
Environment modifier

gili-labs:
Added the environment(keyPath:) method to InspectableView, which gives a unit test the ability to read the value of the environment value specified by the keyPath parameter. This method is useful for unit testing custom view modifiers that modify custom environment values.


gili-labs:
I needed the capability to unit test custom view modifiers that modify custom environment values. This pull request adds support to do this.

nalexn:
Thanks @gili-labs ! Released a new version that includes your PR

PR #95, opened 2021-03-25
Carthage build fix

renep:
I use carthage for my dependencies and this also runs on my build server(s). Here it does not make sense to install swiftlint, just to build the ViewInspector dependency. 
So I updated the build script, that when the build is perform by carthage the swiftlint command check is ignored. When the project is build in Xcode the build will still fail because of a swiftlint.

I think the same problem could occur when using Cocoapods. I do not use Cocoapods, therefore I have not investigated this futher.

### DIFF
--- a/ViewInspector.xcodeproj/project.pbxproj
+++ b/ViewInspector.xcodeproj/project.pbxproj
@@ -822,7 +822,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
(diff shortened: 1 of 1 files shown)


nalexn:
Hey, thank you for the PR! Please let me know how critical this is for you to get it shortly: I haven't planned a new version in the coming couple of weeks, but I can make an exception

renep:
No problem for me. I can use my fork for now, and switch back to your version when you have released a updated.

nalexn:
I ended up converting the error into a warning so that it does not cause issues for CI regardless of the integration method

PR #97, opened 2021-04-06
Added that the alignment for overlay and background can be tested

renep:
I hope that I have done this properly, because I store the alignment value in the `Medium`.

I wanted to avoid a test like this:
```
XCTAssertEqual( try sut.inspect().emptyView().overlayAlignment(), .bottom)
```
So when the `overlay` is inspected also the `alignment` is processed and saved in the `Medium`
So the test now looks like this:
```
XCTAssertEqual(try try sut.inspect().emptyView().overlay().fixedAlignment(), .bottom)
```


nalexn:
@renep I've taken your implementation and reworked it a bit. Ended up introducing a custom inspection attribute for overlay: `alignment()`

Issue #106, opened 2021-05-04
Tapping on a disabled button still invokes its action


alexbasson:
Given the following code:

```swift
Button(action: someAction) {
    Text("a button")
}
.disabled(true)
```

then following test code:

```swift
try view.find(button: "a button").tap()
```

should not invoke `someAction`.

nalexn:
This is now fixed in v0.7.6. Thanks for the ticket and PR!

PR #107, opened 2021-05-04
Calling tap() on a disabled Button does not invoke its action callback

alexbasson:
Resolves issue #106: https://github.com/nalexn/ViewInspector/issues/106

### DIFF
--- a/Sources/ViewInspector/SwiftUI/Button.swift
+++ b/Sources/ViewInspector/SwiftUI/Button.swift
@@ -46,6 +46,10 @@ public extension InspectableView where View == ViewType.Button {
     
     func tap() throws {
         typealias Callback = () -> Void
+        do {
+            if (try self.isDisabled()) { return }
+        } catch {
+        }
         let callback = try Inspector
             .attribute(label: "action", value: content.view, type: Callback.self)
         callback()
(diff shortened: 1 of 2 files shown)


alexbasson:
Hey Alexey, thanks for reviewing the PR!

I'm not sure that throwing makes sense given the use case that inspired this PR. My actual use case is that the button submits a form, and I want to test that tapping the button doesn't submit the form until the form is filled out; here's a simplified version of the test I'm trying to write:

Code:

```swift
Button(action: submitForm) {
    Text("Submit")
}
.disabled($formIsInvalid)
```

Test:

```
func test_whenSomeConditionIsNotYetMet_tappingTheButtonDoesNotInvokeTheAction() {
    XCTAssertTrue(try view.find(button: "Submit").isDisabled())
    try view.find(button: "Submit").tap()

    XCTAssertFalse(formWasSubmitted) // where `formWasSubmitted` comes from some test double or something

    sut.fillOutForm() // actually, access the form views using `find` and set text on them, etc.

    XCTAssertFalse(try view.find(button: "Submit").isDisabled())
    try view.find(button: "Submit").tap()

    XCTAssertTrue(formWasSubmitted)
}
```

As you can see, I don't really want the first `tap()` to throw, I just want it not to invoke the action that submits the form.

nalexn:
Hey, this issue appeared to be much broader:

1. We want to block interaction not only with a Button, but with all other types of controls
2. `disabled(true)` status should propagate for the entire sub-hierarchy, when applied to a container, such as `HStack`
3. There are several reasons why the interaction with a control should not happen: a. `isDisabled()` b. `isHidden()` c. `allowsHitTesting(false)`

I have all of this in place now, and as you suggested, it silently quits. But given the fact there are 3 different reasons for the control to refuse interaction, I tend to revisit it to throw.

Your test can easily tolerate this, either by using `try?` or wrapping the expression in `XCTAssertThrows`

alexbasson:
Sorry for the delayed response. Yes, I agree that given the broader scope, it makes sense to throw. Thanks again for considering it!

Issue #108, opened 2021-05-08
Inspecting Custom View Modifiers Issue


gili-labs:
Consider the following ViewModifier:

```Swift
@available(iOS 13.0, macOS 10.15, tvOS 13.0, *)
private struct InspectableTestModifier: ViewModifier, Inspectable {
    
    @Binding var flag: Bool
    
    var didAppear: ((Self.Body) -> Void)?
    
    func body(content: Self.Content) -> some View {
        HStack {
            content
            Button(
                action: { self.flag.toggle() },
                label: { Text(flag ? "true" : "false") }
            )
        }
        .onAppear { self.didAppear?(self.body(content: content)) }
    }
}
```

And the following test:

```Swift
    func testViewModifierOnFunction() throws {
        @Binding var flag = false
        var sut = InspectableTestModifier(flag: $flag)
        let exp = sut.on(\.didAppear) { body in
            let text1 = try body.find(relation: .child, traversal: .breadthFirst, skipFound: 0) {
                try $0.text().string() == "false"
            }
            let text2 = try body.find(relation: .child, traversal: .breadthFirst, skipFound: 1) {
                try $0.text().string() == "false"
            }
            
            var label1 = try body.hStack()
                .button(1).labelView().text().string()
            var label2 = try body.hStack().modifier(InspectableTestModifier.self).hStack()
                .button(1).labelView().text().string()
            print("\(text1.pathToRoot) = \(label1)")
            print("\(text2.pathToRoot) = \(label2)")
            
            try body.hStack().button(1).tap()
            
            label1 = try body.hStack()
                .button(1).labelView().text().string()
            label2 = try body.hStack().modifier(InspectableTestModifier.self).hStack()
                .button(1).labelView().text().string()
            print("\(text1.pathToRoot) = \(label1)")
            print("\(text2.pathToRoot) = \(label2)")
        }
        ViewHosting.host(view: Circle().modifier(sut))
        wait(for: [exp], timeout: 3.2)
    }
```

Here's a result of running the test:

```
Test Suite 'Selected tests' started at 2021-05-08 15:58:43.465
Test Suite 'ViewInspectorTests.xctest' started at 2021-05-08 15:58:43.466
Test Suite 'InspectionEmissaryTests' started at 2021-05-08 15:58:43.466
Test Case '-[ViewInspectorTests.InspectionEmissaryTests testViewModifierOnFunction]' started.
hStack().button(1).labelView().text() = false
[...]

PR #110, opened 2021-05-15
Inspector emissary for view modifier

gili-labs:
@nalexn I have created this PR with the changes I made to InspectionEmissary to support view modifiers. It works great. However, the `testViewModifierOnFunction` test fails. I left it this way on purpose in order that you can look at the difference between this and the `InspectionEmissaryForViewModifier`. I do not see a difference. I dumped the body of the modifier in the debugger, and they seem identical and cannot ascertain why the ViewInspector would treat them differently.

### DIFF
--- a/Sources/ViewInspector/InspectionEmissary.swift
+++ b/Sources/ViewInspector/InspectionEmissary.swift
@@ -2,55 +2,39 @@ import SwiftUI
 import Combine
 import XCTest
 
+// MARK: - InspectionEmissaryBase
+
 @available(iOS 13.0, macOS 10.15, tvOS 13.0, *)
-public protocol InspectionEmissary: class {
-    
-    associatedtype V: View & Inspectable
-    typealias Inspection = (InspectableView<ViewType.View<V>>) throws -> Void
+public protocol InspectionEmissaryBase: AnyObject {
+
+    associatedtype V
+    associatedtype Body
+    associatedtype InspectedBody
+    typealias Inspection = (InspectedBody) throws -> Void
     
     var notice: PassthroughSubject<UInt, Never> { get }
-    var callbacks: [UInt: (V) -> Void] { get set }
-    
+    var callbacks: [UInt: (Body) -> Void] { get set }
+
     @discardableResult
     func inspect(after delay: TimeInterval,
                  function: String, file: StaticString, line: UInt,
                  _ inspection: @escaping Inspection
     ) -> XCTestExpectation
-    
+
     @discardableResult
     func inspect<P>(onReceive publisher: P,
                     function: String, file: StaticString, line: UInt,
                     _ inspection: @escaping Inspection
     ) -> XCTestExpectation where P: Publisher, P.Failure == Never
+
+    func setup(inspection: @escaping (InspectedBody) throws -> Void,
+               expectation: XCTestExpectation,
+               function: String, file: StaticString, line: UInt)
 }
 
 @available(iOS 13.0, macOS 10.15, tvOS 13.0, *)
-public protocol InspectionEmissaryForViewModifier: class {
-    
-    associatedtype V: ViewModifier & Inspectable
-    typealias Inspection = (InspectableView<ViewType.ClassifiedView>) throws -> Void
-    
-    var notice: PassthroughSubject<UInt, Never> { get }
-    var callbacks: [UInt: (V.Body) -> Void] { get set }
-    
-    @discardableResult
-    func inspect(after delay: TimeInterval,
-                 function: String, file: StaticString, line: UInt,
-                 _ inspection: @escaping Inspection
-    ) -> XCTestExpectation
-    
-    @discardableResult
-    func inspect<P>(onReceive publisher: P,
-                    function: String, file: StaticString, line: UInt,
-                    _ inspection: @escaping Inspection
-    ) -> XCTestExpectation where P: Publisher, P.Failure == Never
-}
+public extension InspectionEmissaryBase {
 
-// MARK: - Default Implementation
-
-@available(iOS 13.0, macOS 10.15, tvOS 13.0, *)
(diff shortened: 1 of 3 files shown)


gili-labs:
Note, I have tried searching for a second text string in the context of the async calls and have found none. As I said, I'm trying ascertain the difference between the `on` function and the use of an `InspectionForViewModifier` object.

nalexn:
At the first glance, I cannot grasp why this happens - really bizarre. I'm about to release a new version - wanted to include this PR as well, but it seems like it deserves further investigation and possibly a revisited approach - I'd ideally come to the `{ self.inspection.visit(self, $0) }` syntax for ViewModifiers, which is more concise and non-recursive. This may as well fix this issue.

nalexn:
Reference to the related discussion: #108

gili-labs:
I have a thought and want to run it by you. The `on` method and `inspection` methods for view modifiers both expect the view modifier's body by calling `self.body(content: content)`. However, I'm thinking this returns a new view created by `body(content:)`, rather than the view originally created when the modified view called `modifier(_:)`. Hence, I'm thinking you're right, we need to figure out how to extract the body from an instance of view modifier. I think this might present a problem as well, because when we create an instance of a view modifier and pass it to `modifier(_:)`, it doesn't create a view, rather SwiftUI at some point calls the instance's `body(content:)` method and encapsulates it in a `ModifiedContent` view.

gili-labs:
@nalexn: I think I solved the problem with dynamic inspection of view extensions. I having a working `on` method, but I to refactor the `inspection` method. Stay tuned.

gili-labs:
@nalexn: Please review the changes I made to this PR. These changes fix the `on` function and the inspector emissary supporting view modifiers.

gili-labs:
@nalexn: Don't merge this yet. I'm having issues defining inspection emissaries in a way that doesn't require the runtime to import ViewInspector.

gili-labs:
The last commit resolves the problem I was experiencing. However, I had to eliminate `InspectionEmissaryBase`, which results in some code duplication. I was trying to avoid this duplication by introducing a base protocol, but this introduced issues that I didn't understand. I'm going to revisit this once there's a version of Swift that [unlocks existentials](https://github.com/apple/swift-evolution/blob/main/proposals/0309-unlock-existential-types-for-all-protocols.md).

nalexn:
Hey - reviewed the code - looks good and works great! The only concern is the requirement to refactor the ViewModifier to return a custom view.

I thought a bit how can we eliminate the recursive `body(content)` call. In our discussion, I referred to a `content` being a stub, which remains a stub in runtime, so an idea came to my mind - what if we don't need to rely on SwiftUI to provide the content parameter and pass in ours? In fact, I already do this for a simple unwrap of [ViewModifiers](https://github.com/nalexn/ViewInspector/blob/master/Sources/ViewInspector/BaseTypes.swift#L36).

So if we add this extension:

```swift
extension ViewModifier {
    func body() -> Any {
        return body(content: _ViewModifier_Content<Self>())
    }
}
```

We'll be able to construct the Body at any moment we want, including when the ViewModifier is hosted on screen and ready for async inspection.

That is, if we previously had

```swift
struct InspectableTestModifier: ViewModifier, Inspectable {
    
    var didAppear: ((Self.Body) -> Void)?
    
    func body(content: Self.Content) -> some View {
        HStack {
            EmptyView()
            content
                .padding(.top, 15)
        }
        .onAppear { self.didAppear?(self.body(content: content)) }
    }
}
```

We can do instead

```swift
struct InspectableTestModifier: ViewModifier, Inspectable {
    
    var didAppear: ((Self) -> Void)?
    
    func body(content: Self.Content) -> some View {
        HStack {
            EmptyView()
            content
                .padding(.top, 15)
        }
        .onAppear { self.didAppear?(self) }
    }
}
```

just like for an arbitrary custom view, but for the `on` function take an additional step and unwrap the content with `body()` call.

I'd want to experiment more with this approach, as it promises a unified approach for Views and ViewModifiers without requiring a refactoring of the ViewModifier's body

gili-labs:
Looks like a good idea.

nalexn:
Have a look at this [slight refactoring](https://github.com/nalexn/ViewInspector/commit/55e1eea1f1c9db59019785774ad94bc2288a269e). The simple tests pass, but I want to add your tests with more sophisticated ViewModifier's structure to assure there are no issues like you've encountered

gili-labs:
Wow, you made it work! I went down this path at one point, but couldn't quite make it do the right thing. By doing this, does one still have to refactor the view out of the body of the view modifier? And you mentioned something above about requiring an extension to ViewModifier. Where do you define that?

nalexn:
> By doing this, does one still have to refactor the view out of the body of the view modifier?

Nah, I really want to avoid it, and so far it works without it. I'm now refactoring the `InspectionEmissary` to remove the `V: View & Inspectable` requirement and make it work for both `View` and `ViewModifier` - this coming along nicely so far and looks promising. I have only 1 hour left today, will try to complete it early this week

nalexn:
Hey @gili-labs , I think I got to a workable solution in this [branch](https://github.com/nalexn/ViewInspector/tree/0.8.0).

nalexn:
@gili-labs I tested it with your examples and it works as expected, I'll likely be preparing a release soon with this code. I think I should close this PR without merging, but I'll give you a shout-out in the release notes as your work played an essential role in solving this challenging task!

gili-labs:
This sounds great! Are you planning on releasing today?

nalexn:
I want to review and merge another PR, plus check if I can put in some more "low hanging fruits". Can easily defer it for a couple of days.

nalexn:
This is resolved now in [v0.8.0](https://github.com/nalexn/ViewInspector/releases/tag/0.8.0)

Issue #59, opened 2020-12-04
How can I inspect a Map?


smartcat32:
```
struct MapSwiftUIView: View {
    
    @ObservedObject var viewModel: MapViewModel
    
    var body: some View {
        Map(coordinateRegion: self.$viewModel.region)
    }
}
```

I'd like to verify the Map is showing the region I gave to the MapViewModel (the center lat/lng and span) in my unit test.
sut.inspect().Map()  // no member 'Map'.  

Are the new SwiftUI views of iOS 14 not supported yet?

nalexn:
Hey @smartcat32 , the `Map` and `MapAnnotation` views have fallen out of my scope because they were defined outside of the SwiftUI framework. I'll add them in one of the coming versions.
I wonder how can I scan all the iOS SDK frameworks for such one-off `Views`...

nalexn:
I figured out that the easiest way to find all SwiftUI - companion frameworks are by looking into the iOS SDK content in Xcode's folder.
Here are the extensions for Xcode 12.2 that can be imported and explored right in the Xcode:

```swift
import _AVKit_SwiftUI
import _AuthenticationServices_SwiftUI
import _HomeKit_SwiftUI
import _MapKit_SwiftUI
import _QuickLook_SwiftUI
import _SceneKit_SwiftUI
import _SpriteKit_SwiftUI
import _StoreKit_SwiftUI
```
I'll update the [reference](https://github.com/nalexn/ViewInspector/blob/master/readiness.md) to reflect the currently not implemented SwiftUI APIs

Tyler-Keith-Thompson:
I've got a [branch](https://github.com/Tyler-Keith-Thompson/ViewInspector/tree/map-support) I'm working on to add support for Maps. Hoping to submit a PR in a couple of days

nalexn:
This is now available in [v0.8.0](https://github.com/nalexn/ViewInspector/releases/tag/0.8.0)

PR #112, opened 2021-05-26
Map support

Tyler-Keith-Thompson:
#59 requests the ability to inspect a SwiftUI Map, this adds that ability but does not yet add the ability to inspect annotations.

I was going to add support for annotations as well before I submitted the PR but annotations are a bit more tricky than I had originally suspected. I figured it might be better to just have 2 PRs, one for supporting the map and one for annotations.


nalexn:
Thanks for contributing @Tyler-Keith-Thompson ! I'll merge this in as soon as I get time. Good job!

PR #120, opened 2021-06-17
Improve NavigationLinkTests as documentation

jonreid:
These tests worked to test your code. But the use of a `Text` view as the navigation destination was confusing, and prevented the tests from serving as examples.

In this PR, I introduce a `TestView` as the destination for every `NavigationLink` in these tests.

### DIFF
--- a/Tests/ViewInspectorTests/SwiftUI/NavigationLinkTests.swift
+++ b/Tests/ViewInspectorTests/SwiftUI/NavigationLinkTests.swift
@@ -7,22 +7,22 @@ final class NavigationLinkTests: XCTestCase {
     
     func testEnclosedView() throws {
         let view = NavigationLink(
-            destination: Text("Screen 1")) { Text("GoTo 1") }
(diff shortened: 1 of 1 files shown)


Tests/ViewInspectorTests/SwiftUI/NavigationLinkTests.swift
```diff
@@ -7,39 +7,39 @@ final class NavigationLinkTests: XCTestCase {
     
     func testEnclosedView() throws {
         let view = NavigationLink(
-            destination: Text("Screen 1")) { Text("GoTo 1") }
-        let text = try view.inspect().navigationLink().text().string()
-        XCTAssertEqual(text, "Screen 1")
+            destination: TestView(parameter: "Screen 1")) { Text("GoTo 1") }
+        let nextView = try view.inspect().navigationLink().view(TestView.self).actualView()
```
jonreid, line 11: This is the "missing piece" you had to explain to me.

PR #124, opened 2021-08-03
Include ToolBarItem in readiness.md

ptrkstr:
Added in iOS 14.
Link to docs: https://developer.apple.com/documentation/swiftui/toolbaritem#


nalexn:
Thank you @patrickbdev , I bet there are more APIs from iOS 14 that are missing, I'll need to find a way to scan those somehow.

PR #150, opened 2021-12-13
fix so that environmentObjects are injected into .actualView()

### LINKED ISSUE #149: How to verify that a view injects an environmentObject to one of its subviews
(A solution proposal ready become a PR for this issue, if you would be willing to consider it)

I have a custom view that uses custom child views. The child views depends on an @EnvironmentObject.
```swift
struct ParentView: View {
   var body: some View {
       ChildView()
            .environmentObject(theObject) // the code I want to verify
   }
}

struct ChildView: View {
   @EnvironmentObject var anObject: TheObject
   var body: some View {
      EmptyView()
   }
}

class TheObject: ObservableObject {}
```
I want to test that the parent view injects the environment object as expected, but when I try:

```swift
       let expectation = XCTestExpectation()
       var sut = TestCustomView()
       sut.on(\.didAppear) { view in
          defer { expectation.fulfill() }
          _ = try view.find(ChildView.self).actualView().anObject
       }
       ViewHosting.host(view: sut.environmentObject(TheObject()))
       wait(for: [expectation], timeout: 1)

```
I get fatal error that anObject is not set on the child. As I understand other issues concerning EnvironmentObject (#88, #32, #62) this is not (yet?) supported. 

So, then I thought, well maybe I can at least check that the parent view has the modifier set with the type of object that I expect, but I could not find any support for that. I found support for getting the modifier-injected-value when it was writable-key.

After some debugging/code broswing/reading, I found that there the code was quite prepared for what I wwas looking for and there was only a few funcs to be added to make it work. And, low and behold, it seems to work.

Would you be interesting in a PR (including tests) for this?
Or, would you recommend another way of verifying the behaviour concerning environmentObject?

pobengtsson:
PR that fixes #149 

Appreciate your review and suggestions for improvements.

### DIFF
--- a/Sources/ViewInspector/SwiftUI/CustomView.swift
+++ b/Sources/ViewInspector/SwiftUI/CustomView.swift
@@ -101,7 +101,11 @@ public extension InspectableView where View: MultipleViewContent {
 public extension InspectableView where View: CustomViewType {
     
     func actualView() throws -> View.T {
-        return try Inspector.cast(value: content.view, type: View.T.self)
+       var view = try Inspector.cast(value: content.view, type: View.T.self)
+       content.medium.environmentObjects.forEach {
+          view.inject(environmentObject: $0)
+       }
+       return view
     }
 }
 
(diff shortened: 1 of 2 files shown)


nalexn:
Yep, that's exactly how I would do it, good job!

PR #184, opened 2022-07-20
Implement callOnSubmit

babbage:
With iOS 15, macOS 12, and related updates, SwiftUI added an `onSubmit` method to add an action to perform when the user submits a value to this view. ([onSubmit(of:_:) documentation](https://developer.apple.com/documentation/swiftui/texteditor/onsubmit(of:_:)))

This PR adds support to call `onSubmit` from ViewInspector, and thus test the `onSubmit` method has been correctly implemented for a view.

`onSubmit` does not require any parameters (other than the action to call) but does provide the option of limiting an `onSubmit` method to a specific `SubmitTriggers` instance, such as `.text` or `.search`, to limit the scope of an `onSubmit` modifier as these can be applied to an entire view hierarchy. The formal test here is only tested this with `TextField` at this point. I demonstrated in other testing (not in the PR) that the method I have implemented here for ViewInspector of `callOnSubmit` will call these trigger-specific `onSubmit` methods set on a view, as well as one set without an explicit submit trigger. The current implementation here of `callOnSubmit` will call the first defined `onSubmit` action, regardless of what trigger may be specified on it in the view. 

It would be a useful future feature of ViewInspector to be able to specifically emulate a submit event that occurred on these two separate types within a view. I was unable to determine whether this is possible with the current implementation of ViewInspector. Therefore, the current implementation calls any onSubmit handler present on a view. For many testing scenarios, this seems likely to be enough. If guidance can be provided on how to target text- vs. search-triggered submit events, I would be happy to extend this implementation.

In the meantime, I have scratched my own itch of being able to test my `onSubmit` methods.

### DIFF
--- a/Sources/ViewInspector/Modifiers/EventsModifiers.swift
+++ b/Sources/ViewInspector/Modifiers/EventsModifiers.swift
@@ -26,3 +26,16 @@ public extension InspectableView {
         callback(value)
     }
 }
+
+@available(iOS 15, macOS 12, tvOS 15, watchOS 8, *)
+public extension InspectableView {
+
+    func callOnSubmit() throws {
+        let callback = try modifierAttribute(
+            modifierName: "OnSubmitModifier",
+            path: "modifier|action",
+            type: (() -> Void).self,
+            call: "onSubmit")
+        callback()
+    }
+}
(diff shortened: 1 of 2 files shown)


babbage:
If you would like the inclusion of the unit tests that show type-specific `onSubmit` modifiers are also called by `callOnSubmit` I already have those and would be happy to add them to the PR.

nalexn:
Thanks for the PR @babbage , and apologies for the late feedback. I'm happy to merge this in, let me explore the possibility of filtering by the "trigger" at first.

PR #191, opened 2022-09-10
feat: mark public keyword to properties of rotation 3d

dscyrescotti:
In the latest release - `v0.9.1`, the properties of `Rotation3D` are not marked as `public` so that they cannot be accessed in other projects/modules that is reliant on `ViewInspect`. This PR includes the fix of exposing them to external modules by marking with `public` keyword. Thanks.


nalexn:
Good find, thanks!

Issue #203, opened 2022-10-11
Question: why do views need to conform to `Inspectable` to be inspected?


bachand:
This is an amazing library. Wow!

I am learning how this library works and I have a question about the `Inspectable` protocol.

Let's say I have a toy SwiftUI view setup like this:

```swift
import SwiftUI

struct MyScreen: View {
  var body: some View {
    MyTextWrapperView(title: "Hello world")
  }
}

struct MyTextWrapperView: View {
  let title: String

  var body: some View {
    Text(title)
  }
}
```

And then let's say I want to verify that my screen shows the string "Hello world" _somewhere_ at any level of the hierarchy. I'd write that test like this:

```swift
import ViewInspector
import XCTest

final class MyScreenTests: XCTestCase {
  func test_verifyTextPresence() {
    XCTAssertNoThrow(try MyScreen().inspect().find(text: "Hello world"))
  }
}
```

I find that for this to to pass I need to conform both `MyScreen` and `MyTextWrapperView` to `Inspectable`.

```swift
extension MyScreen: Inspectable { }
extension MyTextWrapperView: Inspectable { }
```

When building infrastructure for a large codebase it can be challenging to ensure that all of the expected views conform to `Inspectable`. I was hoping that ViewInspector's `find(text:)` would work out of the box without any view needing to be conformed to `Inspectable`. I don't mind needing to conform `MyScreen` to `Inspectable`; the part that is more difficult to reason about is the need to conform child view types, like `MyTextWrapperView`, used within the hierarchy.

I am trying to understand the reason why the `Inspectable` conformance is necessary. I [see](https://github.com/nalexn/ViewInspector/blob/ac7df67c4e0593470eda90a550f8493a81609745/Sources/ViewInspector/BaseTypes.swift#L6-L9) that the `Inspectable` protocol has one property and one method, and that there's a [default implementation](https://github.com/nalexn/ViewInspector/blob/ac7df67c4e0593470eda90a550f8493a81609745/Sources/ViewInspector/BaseTypes.swift#L21-L35) for each when `Self: View`. However I haven't been able to find any documentation as to why the `Inspectable` protocol is necessary.

Thanks again for filling this huge gap in the SwiftUI ecosystem which such a thoughtfully built library 😄

bachand:
I found a related test in this repository.

[...]

PR #213, opened 2022-11-28
[Do not merge] Exploration into removing `Inspectable` conformance requirement

bachand:
_This PR is easy to review commit by commit, if you want to follow the progress._

In this PR I prototype how we might remove the requirement that all views conform to `Inspectable` in order to be inspected. My hope for this PR is to continue the discussion in https://github.com/nalexn/ViewInspector/issues/203 and align on a path for removing the requirement that all views need to conform to `Inspectable` in order to be inspected.

-----

### Status: ⚠️WIP⚠️
In the first commit I updated an existing test to fail, with the goal of fixing this test with future commits. I'm midway through the refactor, focused on removing the `entity` property and `extractContent(…)` method from the `Inspectable` protocol.

I was able to remove `entity` easily. I'm having trouble removing `extractContent(…)` due to the way it requires refactoring the environment injection code. See https://github.com/nalexn/ViewInspector/pull/213#discussion_r1034169681 for where I'm stuck.

I've renamed `Inspectable` to `InspectableProtocol`. My eventual goal is to remove `InspectableProtocol` entirely in favor of the new `Inspectable` struct.

### DIFF
--- a/Sources/ViewInspector/BaseTypes.swift
+++ b/Sources/ViewInspector/BaseTypes.swift
@@ -8,38 +8,6 @@ public protocol InspectableProtocol {
 }
 
 @available(iOS 13.0, macOS 10.15, tvOS 13.0, *)
-public extension InspectableProtocol where Self: View {
-    
-    func extractContent(environmentObjects: [AnyObject]) throws -> Any {
-        var copy = self
-        environmentObjects.forEach { copy = EnvironmentInjection.inject(environmentObject: $0, into: copy) }
-        let missingObjects = EnvironmentInjection.missingEnvironmentObjects(for: copy)
-        if missingObjects.count > 0 {
-            let view = Inspector.typeName(value: self)
-            throw InspectionError
-                .missingEnvironmentObjects(view: view, objects: missingObjects)
-        }
-        return copy.body
-    }
-}
-
-@available(iOS 13.0, macOS 10.15, tvOS 13.0, *)
-public extension InspectableProtocol where Self: ViewModifier {
-    
-    func extractContent(environmentObjects: [AnyObject]) throws -> Any {
-        var copy = self
-        environmentObjects.forEach { copy = EnvironmentInjection.inject(environmentObject: $0, into: copy) }
-        let missingObjects = EnvironmentInjection.missingEnvironmentObjects(for: copy)
-        if missingObjects.count > 0 {
-            let view = Inspector.typeName(value: self)
-            throw InspectionError
-                .missingEnvironmentObjects(view: view, objects: missingObjects)
-        }
-        return copy.body()
-    }
-}
-
-@available(iOS 13.0, macOS 10.15, tvOS 13.0, *)
 public protocol SingleViewContent {
     static func child(_ content: Content) throws -> Content
 }
--- a/Sources/ViewInspector/BaseTypes.swift
+++ b/Sources/ViewInspector/BaseTypes.swift
@@ -3,12 +3,12 @@ import SwiftUI
 // MARK: - Protocols
 
 @available(iOS 13.0, macOS 10.15, tvOS 13.0, *)
-public protocol Inspectable {
+public protocol InspectableProtocol {
     func extractContent(environmentObjects: [AnyObject]) throws -> Any
 }
 
 @available(iOS 13.0, macOS 10.15, tvOS 13.0, *)
-public extension Inspectable where Self: View {
+public extension InspectableProtocol where Self: View {
     
     func extractContent(environmentObjects: [AnyObject]) throws -> Any {
         var copy = self
@@ -24,7 +24,7 @@ public extension Inspectable where Self: View {
 }
 
 @available(iOS 13.0, macOS 10.15, tvOS 13.0, *)
-public extension Inspectable where Self: ViewModifier {
+public extension InspectableProtocol where Self: ViewModifier {
     
     func extractContent(environmentObjects: [AnyObject]) throws -> Any {
         var copy = self
--- a/Sources/ViewInspector/BaseTypes.swift
+++ b/Sources/ViewInspector/BaseTypes.swift
@@ -4,22 +4,11 @@ import SwiftUI
 
 @available(iOS 13.0, macOS 10.15, tvOS 13.0, *)
 public protocol Inspectable {
-    var entity: Content.InspectableEntity { get }
     func extractContent(environmentObjects: [AnyObject]) throws -> Any
 }
 
 @available(iOS 13.0, macOS 10.15, tvOS 13.0, *)
-public extension Content {
-    enum InspectableEntity {
-        case view
-        case viewModifier
-        case gesture
-    }
-}
-
-@available(iOS 13.0, macOS 10.15, tvOS 13.0, *)
 public extension Inspectable where Self: View {
-    var entity: Content.InspectableEntity { .view }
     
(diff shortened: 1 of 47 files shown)


Sources/ViewInspector/ViewSearchIndex.swift
```diff
@@ -91,19 +91,19 @@ internal extension ViewSearch {
...
+            switch content.view {
+            case _ as any View:
                 return .init(ViewType.View<ViewType.Stub>.self, genericTypeName: name)
-            case .viewModifier:
+            case _ as any ViewModifier:
                 return .init(ViewType.ViewModifier<ViewType.Stub>.self, genericTypeName: name)
-            case .gesture:
+            case _ as any Gesture:
+                break
+            default:
```
bachand, line 104: We add a `default` case but we haven't made this code any less "safe". Previously we would return `nil` if `content.view` did not conform to `Inspectable`.

Sources/ViewInspector/ViewSearch.swift
```diff
@@ -378,13 +378,11 @@ private extension UnwrappedView {
...
                 return "popover"
             }
-            if let inspectable = view as? InspectableProtocol {
-                let missingObjects = inspectable.missingEnvironmentObjects
-                if missingObjects.count > 0 {
-                    return InspectionError
-                        .missingEnvironmentObjects(view: name, objects: missingObjects)
-                        .localizedDescription
-                }
+            let missingObjects = EnvironmentInjection.missingEnvironmentObjects(for: view)
```
bachand, line 381: I'm not sure what the implications are of checking for missing environment objects are in all views instead of only views that conform to `Inspectable`.

Tests/ViewInspectorTests/SwiftUI/CustomViewTests.swift
```diff
@@ -50,12 +50,12 @@ final class CustomViewTests: XCTestCase {
         XCTAssertNoThrow(try view.inspect().view(EnvironmentStateTestView.self))
     }
 
-    func testEnvironmentObjectActualView() throws {
-        let viewModel = ExternalState()
-        let view = EnvironmentStateTestView().environmentObject(viewModel)
-        _ = try view.inspect().view(EnvironmentStateTestView.self)
-            .actualView().viewModel
-    }
+//    func testEnvironmentObjectActualView() throws {
```
bachand, line 53: Commit b5364cf3f93c834cc96ded580ae8267deb7784ec resulted in this test crashes.
Screen Shot 2022-11-28 at 3 59 15 PM


The same commit also resulted in many other tests starting to fail.
Screen Shot 2022-11-28 at 3 57 33 PM

bachand: My goal in this commit is to refactor the environment injections functions to be static functions that operate on a view rather than functions that are called directly on a view. This change is necessary to relax the restriction that all views conform to `Inspectable`.

bachand: @nalexn could you advise me on the WIP changes to EnvironmentInjection.swift? I can't see how I've functionally changed those methods though it's clear from the crashing/failing tests that I've done something to alter the behavior. I'm hoping the bug may be more obvious to someone who has spent more time in this code. Thanks!

Sources/ViewInspector/BaseTypes.swift
```diff
@@ -9,33 +81,17 @@ public protocol InspectableProtocol {
...
-        var copy = self
-        environmentObjects.forEach { copy = EnvironmentInjection.inject(environmentObject: $0, into: copy) }
-        let missingObjects = EnvironmentInjection.missingEnvironmentObjects(for: copy)
-        if missingObjects.count > 0 {
-            let view = Inspector.typeName(value: self)
-            throw InspectionError
-                .missingEnvironmentObjects(view: view, objects: missingObjects)
-        }
-        return copy.body
+        try Inspectable.withView(self).extractContent(environmentObjects: environmentObjects)
```
bachand, line 86: Eventually we could inline this code and remove the `extractContent(…)` method from `InspectableProtocol`, which would get us closer to relaxing the requirement that views conform to `InspectableProtocol` in order to be inspected.

nalexn:
Hey, I've left a [comment](https://github.com/nalexn/ViewInspector/issues/203#issuecomment-1331728044) in the related discussion. If I understand your endeavor correctly, you want to remove the `Inspectable` protocol conformance requirement on the custom views. If that's the case, I'm afraid what you're trying to achieve isn't possible, as I described in the comment. Let me know if I misunderstood anything

Tests/ViewInspectorTests/SwiftUI/CustomViewTests.swift
bachand: I wonder if the issue with my refactor may be related to structs being copied.

bachand: Thanks @erichoracek for helping me resolve the crash in `testEnvironmentObjectActualView()` and many of the failing tests. Eric discovered that I needed to update [these lines](https://github.com/nalexn/ViewInspector/blob/b65b8381e699a4ae3ae82f5806ac5e40b3676b2e/Sources/ViewInspector/EnvironmentInjection.swift#L30-L32) to be `MemoryLayout<T>` instead of `MemoryLayout<Self>`.

bachand: I now see that b65b8381e699a4ae3ae82f5806ac5e40b3676b2e is still causing some tests to fail. I will need to investigate these next.

Test failures before applying b65b8381e699a4ae3ae82f5806ac5e40b3676b2e. It's expected that this test fails since I broke the test intentionally in the first commit.

Screen Shot 2022-12-01 at 2 16 09 PM

Test failures after applying b65b8381e699a4ae3ae82f5806ac5e40b3676b2e.
Screen Shot 2022-12-01 at 2 17 44 PM

Sources/ViewInspector/SwiftUI/CustomView.swift
```diff
@@ -2,15 +2,15 @@ import SwiftUI
 
 @available(iOS 13.0, macOS 10.15, tvOS 13.0, *)
 public protocol CustomViewType {
-    associatedtype T: InspectableProtocol
```
bachand: Now that environment injection is decoupled from `InspectableProtocol`, we no longer need this requirement.

Sources/ViewInspector/SwiftUI/CustomView.swift
```diff
@@ -72,7 +72,7 @@ internal extension Content {
 @available(iOS 13.0, macOS 10.15, tvOS 13.0, *)
 public extension InspectableView where View: SingleViewContent {
     
-    func view<T>(_ type: T.Type) throws -> InspectableView<ViewType.View<T>> where T: InspectableProtocol {
```
bachand: This was necessary because previously `ViewType.View<T>` needed to conform to `InspectableProtocol` though that requirement no longer exists.

bachand:
Closing this PR in favor of https://github.com/nalexn/ViewInspector/pull/216. https://github.com/nalexn/ViewInspector/pull/216 successfully decouples content extraction from the `Inspectable` conformance. In continuing to work on this PR, I realized that I could break off that chunk into a standalone change that wouldn't affect the public API. Allowing views that don't conform to `Inspectable` to be inspected will require additional work.

PR #216, opened 2022-12-19
Mb  inspectable is marker protocol

bachand:


Sources/ViewInspector/ViewSearchIndex.swift
```diff
@@ -91,19 +91,19 @@ internal extension ViewSearch {
...
+            switch content.view {
+            case _ as any View:
                 return .init(ViewType.View<ViewType.Stub>.self, genericTypeName: name)
-            case .viewModifier:
+            case _ as any ViewModifier:
                 return .init(ViewType.ViewModifier<ViewType.Stub>.self, genericTypeName: name)
-            case .gesture:
+            case _ as any Gesture:
+                break
+            default:
```
bachand, line 104: We add a `default` case but we haven't made this code any less "safe". Previously we would return `nil` if `content.view` did not conform to `Inspectable`.

Sources/ViewInspector/SwiftUI/CustomView.swift
```diff
@@ -2,15 +2,15 @@ import SwiftUI
 
 @available(iOS 13.0, macOS 10.15, tvOS 13.0, *)
 public protocol CustomViewType {
-    associatedtype T: Inspectable
```
bachand, line 5: Now that environment injection is decoupled from `InspectableProtocol`, we no longer need this requirement.

Sources/ViewInspector/SwiftUI/CustomView.swift
```diff
@@ -72,7 +72,7 @@ internal extension Content {
 @available(iOS 13.0, macOS 10.15, tvOS 13.0, *)
 public extension InspectableView where View: SingleViewContent {
     
-    func view<T>(_ type: T.Type) throws -> InspectableView<ViewType.View<T>> where T: Inspectable {
```
bachand, line 75: This was necessary because previously `ViewType.View<T>` needed to conform to `InspectableProtocol` though that requirement no longer exists.

Sources/ViewInspector/EnvironmentInjection.swift
```diff
@@ -3,10 +3,11 @@ import SwiftUI
 // MARK: - EnvironmentObject injection
 
 @available(iOS 13.0, macOS 10.15, tvOS 13.0, *)
-internal extension Inspectable {
-    var missingEnvironmentObjects: [String] {
+
+internal enum EnvironmentInjection {
+    static func missingEnvironmentObjects(for entity: Any) -> [String] {
```
bachand, line 8: I'm open to better names than `entity`.

Sources/ViewInspector/BaseTypes.swift
```diff
@@ -7,38 +7,6 @@ public protocol Inspectable {
     func extractContent(environmentObjects: [AnyObject]) throws -> Any
 }
 
-@available(iOS 13.0, macOS 10.15, tvOS 13.0, *)
```
bachand, line 11: In the `Move code without changes` commit I consolidate the extraction code into a single file.

Sources/ViewInspector/SwiftUI/CustomView.swift
```diff
@@ -55,14 +55,16 @@ internal extension Content {
     
     func extractCustomView() throws -> Content {
         let inspectable = try Inspector.cast(value: view, type: Inspectable.self)
```
bachand, line 57: It's possible to remove this line now and not cast to `Inspectable`. In this PR we leave the requirement that views conform to `Inspectable` to be inspected.

Sources/ViewInspector/SwiftUI/CustomView.swift
```diff
@@ -126,21 +128,6 @@ public extension NSViewControllerRepresentable where Self: Inspectable {
     }
 }
 
-@available(iOS 13.0, macOS 10.15, tvOS 13.0, *)
```
bachand, line 129: These conformances are no longer necessary since we threw these errors in `ContentExtractor`.

https://github.com/nalexn/ViewInspector/blob/47d11e623ddd0477115df60aaa81fdd9c953c9cf/Sources/ViewInspector/ContentExtraction.swift#L41-L70

Package.swift
```diff
@@ -1,4 +1,4 @@
-// swift-tools-version:5.3
+// swift-tools-version:5.7
```
bachand, line 1: We need to target Swift 5.7 to be able to access this functionality: https://github.com/apple/swift-evolution/blob/main/proposals/0309-unlock-existential-types-for-all-protocols.md

bachand:
@nalexn I'm curious to see what you think of this PR. I'm very open to making revisions (e.g. adding tests, updating naming, restructuring). Lmk!

nalexn:
I've continued the process of de-Inspectable-fication in [this branch](https://github.com/nalexn/ViewInspector/tree/bachand-mb--inspectable-is-marker-protocol). Haven't reviewed the original code in the PR yet, just continued building on top of it. There is no doubt this PR will be a huge milestone for the library, yet, there is quite some work left to make `Inspectable` conformance a true legacy that can easily be removed. Too many APIs have this restriction, and when I remove them, they start to collide with more loose APIs. I'll continue exploration

bachand:
> I've continued the process of de-Inspectable-fication in [this branch](https://github.com/nalexn/ViewInspector/tree/bachand-mb--inspectable-is-marker-protocol).

I checked out the branch. I see you added a protocol called `SwiftUICitizen`, and that you conformed SwiftUI library views to this protocol. I'm curious to learn more about why you added this protocol and how you intend to use it (so I can follow along!).

> Haven't reviewed the original code in the PR yet, just continued building on top of it.

👍

> There is no doubt this PR will be a huge milestone for the library, yet, there is quite some work left to make `Inspectable` conformance a true legacy that can easily be removed.

Agreed. I felt this as well as I tried to push further when I was working on this PR.

> Too many APIs have this restriction, and when I remove the restriction, they start to collide with more loose APIs. I'll continue exploration

Makes sense. Keep me in the loop about how I can be helpful!

nalexn:
> I'm curious to learn more about why you added this protocol and its purpose internal to the library (so that I can follow along!).

The library was using `Inspectable` not only for cracking the custom views but also for distinguishing them from the standard SwiftUI views. As I started removing `Inspectable` from APIs things started falling apart, for example, `find` functions would always appeal to ContextExtractor because it'd now think all Views are custom.

Since we've removed a distinctive mark from custom views, this now had to be added to standard views, thus SwiftUICitizen mark. I was also hoping to use this for maintaining the `inspect` APIs, which have different overloads for custom vs standard views, and I got this working with one exception - `extension ModifiedContent: SwiftUICitizen`. For some reason compiler continues to not notice the conformance when selecting the `inspect` overload among a few, and this drives me nuts. I tend to think we'll need to make breaking API changes for this version and remove / deprecate a few `inspect` overloads. This may be the best timing to do this actually, as there are requests to support async inspection, which would mean the introduction of even more overloads.

nalexn:
I've completed `Inspectable` protocol deprecation. This is now part of [0.9.3](https://github.com/nalexn/ViewInspector/tree/0.9.3) release branch. @bachand thanks for this PR!

bachand:
Thanks @nalexn for finishing the work and for publishing the 0.9.3 release 🙏

PR #228, opened 2023-02-08
Add inspection support for Layout protocol based views

### LINKED ISSUE #226: Cannot find subviews of a custom `Layout`
I have noticed that ViewInspector cannot find subviews when I use a [custom layout](https://developer.apple.com/documentation/swiftui/layout).

I created a minimal repro. Below is a simple screen powered by a custom `SimpleHStack` layout that I found online.

```swift
import SwiftUI

struct MyFirstLayoutScreen: View {

  var body: some View {
    if #available(iOS 16.0, *) {
      SimpleHStack {
        Text("hello world 1")
        Text("hello world 2")
        Text("hello world 3")
      }
    } else {
      EmptyView()
    }
  }

}

@available(iOS 16.0, *)
// copied from https://swiftui-lab.com/layout-protocol-part-1/#layout-cache
struct SimpleHStack: Layout {
  struct CacheData {
    var maxHeight: CGFloat
    var spaces: [CGFloat]
  }

  var spacing: CGFloat? = nil

  func makeCache(subviews: Subviews) -> CacheData {
    return CacheData(maxHeight: computeMaxHeight(subviews: subviews),
                     spaces: computeSpaces(subviews: subviews))
  }

  func updateCache(_ cache: inout CacheData, subviews: Subviews) {
    cache.maxHeight = computeMaxHeight(subviews: subviews)
    cache.spaces = computeSpaces(subviews: subviews)
  }

  func sizeThatFits(proposal: ProposedViewSize, subviews: Subviews, cache: inout CacheData) -> CGSize {
    let idealViewSizes = subviews.map { $0.sizeThatFits(.unspecified) }
    let accumulatedWidths = idealViewSizes.reduce(0) { $0 + $1.width }
    let accumulatedSpaces = cache.spaces.reduce(0) { $0 + $1 }

    return CGSize(width: accumulatedSpaces + accumulatedWidths,
                  height: cache.maxHeight)
  }

  func placeSubviews(in bounds: CGRect, proposal: ProposedViewSize, subviews: Subviews, cache: inout CacheData) {
    var pt = CGPoint(x: bounds.minX, y: bounds.minY)

    for idx in subviews.indices {
      subviews[idx].place(at: pt, anchor: .topLeading, proposal: .unspecified)

      if idx < subviews.count - 1 {
        pt.x += subviews[idx].sizeThatFits(.unspecified).width + cache.spaces[idx]
      }
    }
  }

  func computeSpaces(subviews: LayoutSubviews) -> [CGFloat] {
    if let spacing {
      return Array<CGFloat>(repeating: spacing, count: subviews.count - 1)
    } else {
      return subviews.indices.map { idx in
        guard idx < subviews.count - 1 else { return CGFloat(0) }

        return subviews[idx].spacing.distance(to: subviews[idx+1].spacing, along: .horizontal)
      }
    }
  }

  func computeMaxHeight(subviews: LayoutSubviews) -> CGFloat {
    return subviews.map { $0.sizeThatFits(.unspecified) }.reduce(0) { max($0, $1.height) }
  }
}
```

This code produces the following screen.

Screenshot 2023-01-31 at 2 38 39 PM

However, the following test does not pass.

```swift
import ViewInspector
import XCTest

final class MyFirstLayoutScreenTests: XCTestCase {

  func test_findsText() {
    XCTAssertNoThrow(
      try MyFirstLayoutScreen().inspect().find(text: "hello world 1"))
  }
}
```

Screenshot 2023-01-31 at 2 39 29 PM

If I change the screen to use the builtin `HStack` instead of the custom `SimpleHStack` the test does pass.

```swift
struct MyFirstLayoutScreen: View {
  
  var body: some View {
    HStack {
      Text("hello world 1")
      Text("hello world 2")
      Text("hello world 3")
    }
  }

}
```

This leads me to believe that ViewInspector does not support finding subviews of custom layouts out of the box. Ideally ViewInspector could automatically find subviews in custom layouts.

I am able to help contribute a fix. Before I dig into the code I'm seeking guidance from @nalexn the library author.

rafael-assis:
## Changes

While trying to inspect SwiftUI Views use the `Layout` [protocol](https://developer.apple.com/documentation/swiftui/layout), it was found out that the content views were not able to be retrieved.

Views that use that protocol internally use that `_VariadicView.Tree` where the `root` property of the tree is the layout struct and the `content` property contains the items that will be laid out.

The bottom line of the issue is that ViewInspector's code passes a path to be navigated in order to find the `content` property **assuming that it's a child property** of the `root` property. 

```swift
    static func child(_ content: Content) throws -> Content {
        let view = try Inspector.attribute(path: "root|content", value: content.view)
        return try Inspector.unwrap(view: view, medium: content.medium)
    }
```
That assumption in the current code covers scenarios such as context menu on older versions of iOS.

The `content` property that contains the Views is **actually a sibling** of `root` property as the type structure dump below shows:

```
▿ Content
  ▿ view : Tree<_LayoutRoot<SimpleHStackLayout>, TupleView<(Text, Text, Text)>>
    ▿ root : _LayoutRoot<SimpleHStackLayout>
      ▿ layout : SimpleHStackLayout
        - spacing : nil
    ▿ content : TupleView<(Text, Text, Text)>
      ▿ value : 3 elements
        ▿ .0 : Text
          ▿ storage : Storage
            ▿ anyTextStorage : <LocalizedTextStorage: 0x000060000358b750>: "LayoutScreen text 1"
          - modifiers : 0 elements
        ▿ .1 : Text
          ▿ storage : Storage
            ▿ anyTextStorage : <LocalizedTextStorage: 0x00006000035ec000>: "LayoutScreen text 2"
          - modifiers : 0 elements
        ▿ .2 : Text
          ▿ storage : Storage
            ▿ anyTextStorage : <LocalizedTextStorage: 0x00006000035ec050>: "LayoutScreen text 3"
          - modifiers : 0 elements
  ▿ medium : Medium
    - viewModifiers : 0 elements
    - transitiveViewModifiers : 0 elements
    - environmentModifiers : 0 elements
    - environmentObjects : 0 elements
```

This change adds a second route in case the `root|content` is not available, it attempts the sibling `content` property which will cover the new scenario.

As the `Layout` based views also `_VariadicView.Tree` under the hood, this fixes #226 created by @bachand.

@nalexn as an FYI: @bachand and I work together at Airbnb and we're using the ViewInspector library in SwiftUI tests.

## Testing

Added tests to exercise the `Layout` based Views on iOS 16 and above.

## Code Coverage results for TreeView.swift file on iOS 16.2:

**Before:**

test_coverage_before

**After:**

coverage_after

## Review
@bachand 
@nalexn

### DIFF
--- a/Sources/ViewInspector/SwiftUI/TreeView.swift
+++ b/Sources/ViewInspector/SwiftUI/TreeView.swift
@@ -9,7 +9,15 @@ internal extension ViewType {
 extension ViewType.TreeView: SingleViewContent {
 
     static func child(_ content: Content) throws -> Content {
-        let view = try Inspector.attribute(path: "content", value: content.view)
+        let view: Any = try {
+            guard let rootContent = try? Inspector.attribute(path: "root|content", value: content.view) else {
+                // Try to return the `contents` property directly if the tree root does not contain a View.
+                // (e.g. Layout, _VariadicView_MultiViewRoot, _VariadicView_UnaryViewRoot)
+                return try Inspector.attribute(path: "content", value: content.view)
+            }
+
+            return rootContent
+        }()
         return try Inspector.unwrap(view: view, medium: content.medium)
     }
 }
--- a/Sources/ViewInspector/SwiftUI/TreeView.swift
+++ b/Sources/ViewInspector/SwiftUI/TreeView.swift
@@ -9,7 +9,7 @@ internal extension ViewType {
 extension ViewType.TreeView: SingleViewContent {
 
     static func child(_ content: Content) throws -> Content {
-        let view = try Inspector.attribute(path: "root|content", value: content.view)
+        let view = try Inspector.attribute(path: "content", value: content.view)
         return try Inspector.unwrap(view: view, medium: content.medium)
     }
 }
--- a/Tests/ViewInspectorTests/SwiftUI/TreeViewTests.swift
+++ b/Tests/ViewInspectorTests/SwiftUI/TreeViewTests.swift
@@ -70,9 +70,9 @@ private struct SampleVariadicViewScreen: View {
         } else {
           opacity = 0.5
         }
-      }) {
+      }, label: {
         Text("Execute")
-      }
+      })
     })
   }
 }
@@ -114,7 +114,7 @@ private struct SimpleHStackLayout: Layout {
     var spaces: [CGFloat]
   }
 
-  var spacing: CGFloat? = nil
+  var spacing: CGFloat?
 
   func makeCache(subviews: Subviews) -> CacheData {
     CacheData(
--- a/Tests/ViewInspectorTests/SwiftUI/TreeViewTests.swift
+++ b/Tests/ViewInspectorTests/SwiftUI/TreeViewTests.swift
@@ -30,7 +30,7 @@ final class TreeViewTests: XCTestCase {
     }
 
   func testVariadicViewTree() throws {
-    let button = try VariadicViewScreen().inspect().find(button: "Execute")
+    let button = try VariadicViewScreen().inspect().find(button: "Change opacity")
     XCTAssertNotNil(button)
   }
 
@@ -62,7 +62,7 @@ private struct VariadicViewScreen: View {
   var body: some View {
     _VariadicView.Tree(VariadicViewScreenOpacityRoot(opacity: opacity),
                        content: {
-      Text("Click the button to execute the action.")
+      Text("Click the button to change the opacity.")
       Button(action: {
         if opacity < 1.0 {
           opacity = 1.0
@@ -70,7 +70,7 @@ private struct VariadicViewScreen: View {
           opacity = 0.5
         }
       }, label: {
-        Text("Execute")
+        Text("Change opacity")
       })
     })
   }
--- a/Tests/ViewInspectorTests/SwiftUI/TreeViewTests.swift
+++ b/Tests/ViewInspectorTests/SwiftUI/TreeViewTests.swift
@@ -28,6 +28,17 @@ final class TreeViewTests: XCTestCase {
         }
         XCTAssertEqual(sut.content.medium.viewModifiers.count, count)
     }
+
+  @available(iOS 16.0, *)
+  func testVariadicViewTree() throws {
+    let button = try SampleVariadicViewScreen().inspect().find(button: "Execute")
+    XCTAssertNotNil(button)
+  }
+
+  @available(iOS 16.0, *)
+  func testLayoutViewTree() throws {
+    XCTAssertNoThrow(try SampleLayoutScreen().inspect().find(text: "SampleLayoutScreen text 2"))
+  }
 }
 
 // MARK: - View Modifiers
@@ -41,4 +52,116 @@ final class GlobalModifiersForTreeView: XCTestCase {
         let sut = EmptyView().contextMenu(ContextMenu(menuItems: { Text("") }))
         XCTAssertNoThrow(try sut.inspect().emptyView())
     }
+}
+
+// MARK: - SampleVariadicViewScreen
+
+@available(iOS 13.0, macOS 10.15, *)
+private struct SampleVariadicViewScreen: View {
+  @State var opacity: Double = 1.0
+
+  var body: some View {
+    _VariadicView.Tree(SampleVariadicViewScreenOpacityRoot(opacity: opacity),
+                       content: {
+      Text("Click the button to execute the action.")
+      Button(action: {
+        if opacity < 1.0 {
+          opacity = 1.0
+        } else {
+          opacity = 0.5
+        }
+      }) {
+        Text("Execute")
+      }
+    })
+  }
+}
+
+// MARK: - SampleVariadicViewScreenRoot
+
+@available(iOS 13.0, macOS 10.15, *)
+private struct SampleVariadicViewScreenOpacityRoot: _VariadicView_MultiViewRoot {
+  let opacity: Double
+
+  func body(children: _VariadicView.Children) -> some View {
+    ForEach(children) { child in
+      child
+        .opacity(opacity)
+    }
+  }
+}
+
+// MARK: - SampleLayoutScreen
+
+@available(iOS 16.0, *)
+private struct SampleLayoutScreen: View {
+  var body: some View {
+    SimpleHStackLayout {
+      Text("SampleLayoutScreen text 1")
+      Text("SampleLayoutScreen text 2")
+      Text("SampleLayoutScreen text 3")
+    }
+  }
+}
+
+// MARK: - SimpleHStack
+
+@available(iOS 16.0, *)
+// copied from https://swiftui-lab.com/layout-protocol-part-1/#layout-cache
+private struct SimpleHStackLayout: Layout {
+  struct CacheData {
+    var maxHeight: CGFloat
+    var spaces: [CGFloat]
+  }
+
+  var spacing: CGFloat? = nil
+
+  func makeCache(subviews: Subviews) -> CacheData {
+    CacheData(
+      maxHeight: computeMaxHeight(subviews: subviews),
+      spaces: computeSpaces(subviews: subviews))
+  }
+
+  func updateCache(_ cache: inout CacheData, subviews: Subviews) {
+    cache.maxHeight = computeMaxHeight(subviews: subviews)
+    cache.spaces = computeSpaces(subviews: subviews)
+  }
+
+  func sizeThatFits(proposal _: ProposedViewSize, subviews: Subviews, cache: inout CacheData) -> CGSize {
+    let idealViewSizes = subviews.map { $0.sizeThatFits(.unspecified) }
+    let accumulatedWidths = idealViewSizes.reduce(0) { $0 + $1.width }
+    let accumulatedSpaces = cache.spaces.reduce(0) { $0 + $1 }
+
+    return CGSize(
+      width: accumulatedSpaces + accumulatedWidths,
+      height: cache.maxHeight)
+  }
+
+  func placeSubviews(in bounds: CGRect, proposal _: ProposedViewSize, subviews: Subviews, cache: inout CacheData) {
+    var pt = CGPoint(x: bounds.minX, y: bounds.minY)
+
+    for idx in subviews.indices {
+      subviews[idx].place(at: pt, anchor: .topLeading, proposal: .unspecified)
+
+      if idx < subviews.count - 1 {
+        pt.x += subviews[idx].sizeThatFits(.unspecified).width + cache.spaces[idx]
+      }
+    }
+  }
+
+  func computeSpaces(subviews: LayoutSubviews) -> [CGFloat] {
+    if let spacing {
+      return [CGFloat](repeating: spacing, count: subviews.count - 1)
+    } else {
+      return subviews.indices.map { idx in
+        guard idx < subviews.count - 1 else { return CGFloat(0) }
+
+        return subviews[idx].spacing.distance(to: subviews[idx + 1].spacing, along: .horizontal)
+      }
+    }
+  }
+
+  func computeMaxHeight(subviews: LayoutSubviews) -> CGFloat {
+    subviews.map { $0.sizeThatFits(.unspecified) }.reduce(0) { max($0, $1.height) }
+  }
 }
--- a/Tests/ViewInspectorTests/SwiftUI/TreeViewTests.swift
+++ b/Tests/ViewInspectorTests/SwiftUI/TreeViewTests.swift
@@ -8,7 +8,7 @@ final class TreeViewTests: XCTestCase {
     
     @available(watchOS, deprecated: 7.0)
     func testEnclosedView() throws {
-        let sut = Text("Test").contextMenu(ContextMenu(menuItems: { Text("Menu") }))
+        let sut = Text("Test").contextMenu(menuItems: { Text("Menu") })
         let text = try sut.inspect().text().string()
         XCTAssertEqual(text, "Test")
     }
@@ -17,7 +17,7 @@ final class TreeViewTests: XCTestCase {
     func testRetainsModifiers() throws {
         let view = Text("Test")
             .padding()
-            .contextMenu(ContextMenu(menuItems: { Text("Menu") }))
+            .contextMenu(menuItems: { Text("Menu") })
             .padding().padding()
         let sut = try view.inspect().text()
         let count: Int
@@ -51,7 +51,7 @@ final class GlobalModifiersForTreeView: XCTestCase {
     
     @available(watchOS, deprecated: 7.0)
     func testContextMenu() throws {
-        let sut = EmptyView().contextMenu(ContextMenu(menuItems: { Text("") }))
+        let sut = EmptyView().contextMenu(menuItems: { Text("") })
         XCTAssertNoThrow(try sut.inspect().emptyView())
     }
 }
--- a/Tests/ViewInspectorTests/SwiftUI/TreeViewTests.swift
+++ b/Tests/ViewInspectorTests/SwiftUI/TreeViewTests.swift
@@ -29,7 +29,7 @@ final class TreeViewTests: XCTestCase {
         XCTAssertEqual(sut.content.medium.viewModifiers.count, count)
     }
 
-  @available(iOS 16.0, *)
+  @available(watchOS, deprecated: 7.0)
   func testVariadicViewTree() throws {
     let button = try SampleVariadicViewScreen().inspect().find(button: "Execute")
     XCTAssertNotNil(button)
--- a/Tests/ViewInspectorTests/SwiftUI/TreeViewTests.swift
+++ b/Tests/ViewInspectorTests/SwiftUI/TreeViewTests.swift
@@ -29,15 +29,14 @@ final class TreeViewTests: XCTestCase {
         XCTAssertEqual(sut.content.medium.viewModifiers.count, count)
     }
 
-  @available(watchOS, deprecated: 7.0)
   func testVariadicViewTree() throws {
-    let button = try SampleVariadicViewScreen().inspect().find(button: "Execute")
+    let button = try VariadicViewScreen().inspect().find(button: "Execute")
     XCTAssertNotNil(button)
   }
 
   @available(iOS 16.0, *)
-  func testLayoutViewTree() throws {
-    XCTAssertNoThrow(try SampleLayoutScreen().inspect().find(text: "SampleLayoutScreen text 2"))
+  func testLayoutBasedViewTree() throws {
+    XCTAssertNoThrow(try LayoutScreen().inspect().find(text: "LayoutScreen text 2"))
   }
 }
 
@@ -54,14 +53,14 @@ final class GlobalModifiersForTreeView: XCTestCase {
     }
 }
 
-// MARK: - SampleVariadicViewScreen
+// MARK: - VariadicViewScreen
 
 @available(iOS 13.0, macOS 10.15, *)
-private struct SampleVariadicViewScreen: View {
+private struct VariadicViewScreen: View {
   @State var opacity: Double = 1.0
 
   var body: some View {
-    _VariadicView.Tree(SampleVariadicViewScreenOpacityRoot(opacity: opacity),
+    _VariadicView.Tree(VariadicViewScreenOpacityRoot(opacity: opacity),
                        content: {
       Text("Click the button to execute the action.")
       Button(action: {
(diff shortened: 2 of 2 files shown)


Tests/ViewInspectorTests/SwiftUI/TreeViewTests.swift
```diff
@@ -28,6 +28,16 @@ final class TreeViewTests: XCTestCase {
         }
         XCTAssertEqual(sut.content.medium.viewModifiers.count, count)
     }
+
+  func testVariadicViewTree() throws {
```
bachand: Nit: we probably want to use 4 spaces of indentation per tab in this repo.

Tests/ViewInspectorTests/SwiftUI/TreeViewTests.swift
```diff
@@ -28,6 +28,16 @@ final class TreeViewTests: XCTestCase {
         }
         XCTAssertEqual(sut.content.medium.viewModifiers.count, count)
     }
+
```
bachand, line 31: I am having trouble understanding how TreeView.swift had 0% code coverage given that the above tests exist.

bachand: Overall I'm wondering if this TreeView.swift file is related to a different type of "tree"? I'm trying to assess if we should be adding these new tests to this or a new file.

Tests/ViewInspectorTests/SwiftUI/TreeViewTests.swift
```diff
@@ -28,6 +28,16 @@ final class TreeViewTests: XCTestCase {
...
+  func testVariadicViewTree() throws {
+    let button = try VariadicViewScreen().inspect().find(button: "Change opacity")
```
bachand: It could be interesting to verify that a child view has the expected opacity applied.

bachand (approved):
Great job figuring this out @rafael-assis ! Let us know what you think @nalexn .

Tests/ViewInspectorTests/SwiftUI/TreeViewTests.swift
rafael-assis: Thanks for catching this! Just pushed a fix to use 4 spaces.

Tests/ViewInspectorTests/SwiftUI/TreeViewTests.swift
rafael-assis: That's a great question! I debugged the existing tests and none of them hit the code in the TreeView.swift file.
I enabled coverage to double check that none of the other tests were exercising this code path.

@nalexn any thoughts this?

nalexn: If you run the tests on earlier os versions (I have 14.4 simulator image installed), that code would get executed for `ContextMenu` in the views. SwiftUI internals get a refactor with every major os version, so we have to keep the old parsers in place to support previous versions.

Sources/ViewInspector/SwiftUI/TreeView.swift
```diff
@@ -9,7 +9,7 @@ internal extension ViewType {
 extension ViewType.TreeView: SingleViewContent {
 
     static func child(_ content: Content) throws -> Content {
-        let view = try Inspector.attribute(path: "root|content", value: content.view)
```
nalexn, line 12: This is not backward compatible with older os versions, see how this is worked around in a [similar case here](https://github.com/nalexn/ViewInspector/blob/4effbd9143ab797eb60d2f32d4265c844c980946/Sources/ViewInspector/Modifiers/CustomStyleModifiers.swift#L15)

Tests/ViewInspectorTests/SwiftUI/TreeViewTests.swift
```diff
@@ -42,3 +52,115 @@ final class GlobalModifiersForTreeView: XCTestCase {
...
 }
+
+// MARK: - VariadicViewScreen
+
+@available(iOS 13.0, macOS 10.15, *)
+private struct VariadicViewScreen: View {
+    @State var opacity: Double = 1.0
+
+    var body: some View {
+        _VariadicView.Tree(VariadicViewScreenOpacityRoot(opacity: opacity),
```
nalexn: I'm curious if you're using `_VariadicView` in the production code? It seems to be the type resolved after [inlining stack views](https://movingparts.io/variadic-views-in-swiftui). I'm ok with this test, though, for the sake of clarity and conciseness, and if it's possible, could we refactor this piece to use a stack view that unfolds into `_VariadicView`?

Tests/ViewInspectorTests/SwiftUI/TreeViewTests.swift
rafael-assis: Thanks for pointing this out @nalexn!

I did run it on iOS 14.5 and noticed that on that version the `Tree` is the `view` property of the content (reason why it covers that code). On iOS 16 the `Tree` only comes in the `ZStack` inside the `menuView` property.

I've also replaced the calls to the deprecated [ContextMenu](https://developer.apple.com/documentation/swiftui/contextmenu) with the direct modifier call to  [contextmenu(menuitems:)](https://developer.apple.com/documentation/swiftui/view/contextmenu(menuitems:)) (also supported since iOS 13) with no major changes in the underlying structure.

Here is the dump for all the variations:

##  iOS 16.2
**Using [contextmenu(menuitems:)]**(https://developer.apple.com/documentation/swiftui/view/contextmenu(menuitems:))
```
▿ Content
  ▿ view : ModifiedContent<Text, ContextMenuModifier<ZStack<ModifiedContent<Text, StyleContextWriter<MenuStyleContext>>>>>
    ▿ content : Text
      ▿ storage : Storage
        ▿ anyTextStorage : <LocalizedTextStorage: 0x0000600000d1a440>: "Test"
      - modifiers : 0 elements
    ▿ modifier : ContextMenuModifier<ZStack<ModifiedContent<Text, StyleContextWriter<MenuStyleContext>>>>
      ▿ _id : IdentityLink
        ▿ _value : ViewIdentity
          - seed : 0
      ▿ _activeMenuID : Environment<ViewIdentity>
        ▿ content : Content
          - keyPath : <WritableKeyPath<EnvironmentValues, ViewIdentity>: 0x600003b2d100>
      ▿ _menuOrder : Environment<MenuOrder>
        ▿ content : Content
          - keyPath : <WritableKeyPath<EnvironmentValues, MenuOrder>: 0x600003b2d180>
      ▿ menuView : ZStack<ModifiedContent<Text, StyleContextWriter<MenuStyleContext>>>
        ▿ _tree : Tree<_ZStackLayout, ModifiedContent<Text, StyleContextWriter<MenuStyleContext>>>
          ▿ root : _ZStackLayout
            ▿ alignment : Alignment
              ▿ horizontal : HorizontalAlignment
                ▿ key : AlignmentKey
                  - bits : 2
              ▿ vertical : VerticalAlignment
                ▿ key : AlignmentKey
                  - bits : 5
          ▿ content : ModifiedContent<Text, StyleContextWriter<MenuStyleContext>>
            ▿ content : Text
              ▿ storage : Storage
                ▿ anyTextStorage : <LocalizedTextStorage: 0x0000600000d34cd0>: "Menu"
              - modifiers : 0 elements
            - modifier : SwiftUI.StyleContextWriter<SwiftUI.MenuStyleContext>()
  ▿ medium : Medium
    - viewModifiers : 0 elements
    - transitiveViewModifiers : 0 elements
    - environmentModifiers : 0 elements
    - environmentObjects : 0 elements

```


**Using the deprecated [ContextMenu](https://developer.apple.com/documentation/swiftui/contextmenu)**
```
▿ Content
  ▿ view : ModifiedContent<Text, ContextMenuModifier<Group<Optional<ZStack<ModifiedContent<Text, StyleContextWriter<MenuStyleContext>>>>>>>
    ▿ content : Text
      ▿ storage : Storage
        ▿ anyTextStorage : <LocalizedTextStorage: 0x00006000013bbd90>: "Test"
      - modifiers : 0 elements
    ▿ modifier : ContextMenuModifier<Group<Optional<ZStack<ModifiedContent<Text, StyleContextWriter<MenuStyleContext>>>>>>
      ▿ _id : IdentityLink
        ▿ _value : ViewIdentity
          - seed : 0
      ▿ _activeMenuID : Environment<ViewIdentity>
        ▿ content : Content
          - keyPath : <WritableKeyPath<EnvironmentValues, ViewIdentity>: 0x6000025f6600>
      ▿ _menuOrder : Environment<MenuOrder>
        ▿ content : Content
          - keyPath : <WritableKeyPath<EnvironmentValues, MenuOrder>: 0x6000025f66c0>
      ▿ menuView : Group<Optional<ZStack<ModifiedContent<Text, StyleContextWriter<MenuStyleContext>>>>>
        ▿ content : Optional<ZStack<ModifiedContent<Text, StyleContextWriter<MenuStyleContext>>>>
          ▿ some : ZStack<ModifiedContent<Text, StyleContextWriter<MenuStyleContext>>>
            ▿ _tree : Tree<_ZStackLayout, ModifiedContent<Text, StyleContextWriter<MenuStyleContext>>>
              ▿ root : _ZStackLayout
                ▿ alignment : Alignment
                  ▿ horizontal : HorizontalAlignment
                    ▿ key : AlignmentKey
                      - bits : 2
                  ▿ vertical : VerticalAlignment
                    ▿ key : AlignmentKey
                      - bits : 5
              ▿ content : ModifiedContent<Text, StyleContextWriter<MenuStyleContext>>
                ▿ content : Text
                  ▿ storage : Storage
                    ▿ anyTextStorage : <LocalizedTextStorage: 0x00006000013bbde0>: "Menu"
                  - modifiers : 0 elements
                - modifier : SwiftUI.StyleContextWriter<SwiftUI.MenuStyleContext>()
  ▿ medium : Medium
    - viewModifiers : 0 elements
    - transitiveViewModifiers : 0 elements
    - environmentModifiers : 0 elements
    - environmentObjects : 0 elements
```

##  iOS 14.5
**Using [contextmenu(menuitems:)]**(https://developer.apple.com/documentation/swiftui/view/contextmenu(menuitems:))
```
▿ Content
  ▿ view : Tree<Container<Text>, ModifiedContent<Text, StyleContextWriter<MenuContext>>>
    ▿ root : Container<Text>
      ▿ content : Text
        ▿ storage : Storage
          ▿ anyTextStorage : <LocalizedTextStorage: 0x7fe457d23b30>
        - modifiers : 0 elements
      ▿ contextMenu : Optional<_ContextMenuContainer>
        - some : SwiftUI._ContextMenuContainer()
    ▿ content : ModifiedContent<Text, StyleContextWriter<MenuContext>>
      ▿ content : Text
        ▿ storage : Storage
          ▿ anyTextStorage : <LocalizedTextStorage: 0x7fe457d212e0>
        - modifiers : 0 elements
      - modifier : SwiftUI.StyleContextWriter<SwiftUI.MenuContext>()
  ▿ medium : Medium
    - viewModifiers : 0 elements
    - transitiveViewModifiers : 0 elements
    - environmentModifiers : 0 elements
    - environmentObjects : 0 elements
```

Using the deprecated [ContextMenu](https://developer.apple.com/documentation/swiftui/contextmenu)
```
▿ Content
  ▿ view : Tree<Container<Text>, Optional<ModifiedContent<AnyView, StyleContextWriter<MenuContext>>>>
    ▿ root : Container<Text>
      ▿ content : Text
        ▿ storage : Storage
          ▿ anyTextStorage : <LocalizedTextStorage: 0x7f8f79522820>
        - modifiers : 0 elements
      ▿ contextMenu : Optional<_ContextMenuContainer>
        - some : SwiftUI._ContextMenuContainer()
    ▿ content : Optional<ModifiedContent<AnyView, StyleContextWriter<MenuContext>>>
      ▿ some : ModifiedContent<AnyView, StyleContextWriter<MenuContext>>
        ▿ content : AnyView
          ▿ storage : <AnyViewStorage<ModifiedContent<Text, StyleContextWriter<MenuContext>>>: 0x7f8f79522f50>
        - modifier : SwiftUI.StyleContextWriter<SwiftUI.MenuContext>()
  ▿ medium : Medium
    - viewModifiers : 0 elements
    - transitiveViewModifiers : 0 elements
    - environmentModifiers : 0 elements
    - environmentObjects : 0 elements
```

Tests/ViewInspectorTests/SwiftUI/TreeViewTests.swift
rafael-assis: Yes. We do have a custom View that calls the `_VariadicView.Tree` directly from its `body` property implementation.

The goal I had in mind for this test is exercising the direct call. The Layout test covers the case of a `View` that unfolds into a `_VariadicView.Tree`.

I'm also okay removing this test if you prefer.

rafael-assis: One thing I'm having a hard time understanding is that modifying the test to the following code, I get a `VStack` with an underlying `_VariadicView.Tree`, but the code  in TreeView.swift is not called. And the button is found.

```swift
    func testVariadicViewTree() throws {
        let button = try VStack {
            Text("Click the button to change the opacity.")
            Button(action: {
                print("test")
            }, label: {
                Text("Change opacity")
            })
        }.inspect().find(button: "Change opacity")
        XCTAssertNotNil(button)
    }
```

```
▿ Content
  ▿ view : VStack<TupleView<(Text, Button<Text>)>>
    ▿ _tree : Tree<_VStackLayout, TupleView<(Text, Button<Text>)>>
      ▿ root : _VStackLayout
        ▿ alignment : HorizontalAlignment
          ▿ key : AlignmentKey
            - bits : 5381443304
        - spacing : nil
      ▿ content : TupleView<(Text, Button<Text>)>
        ▿ value : 2 elements
          ▿ .0 : Text
            ▿ storage : Storage
              ▿ anyTextStorage : <LocalizedTextStorage: 0x7ff7e20c2320>
            - modifiers : 0 elements
          ▿ .1 : Button<Text>
            ▿ _label : Text
              ▿ storage : Storage
                ▿ anyTextStorage : <LocalizedTextStorage: 0x7ff7e20c4a50>
              - modifiers : 0 elements
            - action : (Function)
  ▿ medium : Medium
    - viewModifiers : 0 elements
    - transitiveViewModifiers : 0 elements
    - environmentModifiers : 0 elements
    - environmentObjects : 0 elements
```

Sources/ViewInspector/SwiftUI/TreeView.swift
rafael-assis: Thanks! I added a workaround similar to the one you pointed out. Tests now pass on both iOS 14.5 and 16.2.

One question I had while looking at the `testEnclosedView()` test below: 
**Is there a way to inspect the menu items and find the `Text("Menu")`?**

https://github.com/nalexn/ViewInspector/blob/7a45551ab00f0ff9c8a0996374edcba73705438b/Tests/ViewInspectorTests/SwiftUI/TreeViewTests.swift#L9-L14

Tests/ViewInspectorTests/SwiftUI/TreeViewTests.swift
rafael-assis: The blocking issue I see is that the opacity is applied by the `root` of the `Tree` - `VariadicViewScreenOpacityRoot`. And the test code does not call into it. It only calls the `content` property.

rafael-assis (commented):
Thank you @nalexn and @bachand for reviewing.
I have made some changes experimented with other ones described in the comments.
Please let me know your thoughts.
Thank you!

Tests/ViewInspectorTests/SwiftUI/TreeViewTests.swift
bachand: Based on [this](https://movingparts.io/variadic-views-in-swiftui) article @rafael-assis I thought that the opacity would be applied to each subview when the root is `_Variadic­View_Multi­View­Root` like we have here. However, I think an issue is that there's no way to read the opacity from the subviews.

Sources/ViewInspector/SwiftUI/TreeView.swift
```diff
@@ -9,7 +9,15 @@ internal extension ViewType {
...
+        let view: Any = try {
+            guard let rootContent = try? Inspector.attribute(path: "root|content", value: content.view) else {
+                // Try to return the `contents` property directly if the tree root does not contain a View.
+                // (e.g. Layout, _VariadicView_MultiViewRoot, _VariadicView_UnaryViewRoot)
+                return try Inspector.attribute(path: "content", value: content.view)
+            }
```
bachand, line 17: It could be nice if we had a comment indicating that one version exists for older SwiftUI versions (with the version number). Then the code can be cleaned up if the library stops supporting older SwiftUI versions.

Tests/ViewInspectorTests/SwiftUI/TreeViewTests.swift
rafael-assis: It's correct that the `_Variadic­View_Multi­View­Root` will apply the opacity to each view. But I believe that this will happen in the context of the SwiftUI layout pass. In the context of ViewInspector test the `VariadicViewScreen`'s `body` property is called but the `func body(children: _VariadicView.Children) -> some View` of the `VariadicViewScreenOpacityRoot` is never called.

Sources/ViewInspector/SwiftUI/TreeView.swift
nalexn: True, but sometimes it's hard to tell at which exact os version the change took place - you'd need to do a "bisect" on a wide range of iOS versions, and have them all installed

Tests/ViewInspectorTests/SwiftUI/TreeViewTests.swift
nalexn: I can have a closer look myself at some point, but I have a feeling we might need to explicitly support `_Variadic­View.Multi­View­Root` by defining a `ViewType`, like for `ForEach`, and similarly mark it as `MultipleViewContent`. Then we'll have full control on children extraction, including ability to maintain `opacity` modifiers. Usually modifiers are reset like so: `content.medium.resettingViewModifiers()`

As a side note, you can use `Inspector.print(view) as AnyObject` instead of `dump` when printing view's internals, it's doing a bit extra work when digging into the structure.

Sources/ViewInspector/SwiftUI/TreeView.swift
nalexn: When I print the `sut` with `Inspector.print`, I can see `"Menu"` string in the structure, which is a good sign - searching / direct inspection should be possible

rafael-assis: Yes! I agree!
I created #229 to track this work as it's out of the scope of this PR and the issue it's fixing.

rafael-assis (commented):
@nalexn I have removed the test that uses `_VariadicView` directly.

@bachand and I believe this PR is ready to be merged in order to address #226.
Let us know if you have any additional feedback or concerns.

Thanks again!

Tests/ViewInspectorTests/SwiftUI/TreeViewTests.swift
rafael-assis: It makes a lot of sense.
This seems to be beyond the scope of what we need right now as we're only looking to find the views that are contained in a Layout based view.
But it would definitely be great to have full support for variadic views in a way that the root of the tree is called to apply all the modifiers to the content views.

Tests/ViewInspectorTests/SwiftUI/TreeViewTests.swift
rafael-assis: FYI: I have removed the test that explicitly uses `_VariadicView.Tree` so it does not cause confusion on what the code is supposed to cover.

bachand (approved):
Thanks @nalexn for your help, and great work figuring this out @rafael-assis ! We're excited for an updated version of `ViewInspector` that includes this fix. We are currently blocked from inspecting elements inside some widely used views that leverage trees.

bachand (approved):

nalexn:
Thank you both for the collaboration. I've merged the PR into the active 0.9.6 branch. I'm curious if you can use the branch reference instead of the release tag for some time in your project? I try to bundle a few fixes or features per each release (unless there is a hotfix for a pressing issue), but so far there is just this PR, so I would ideally release later when more work is accumulated. Please let me know!

bachand:
> I've merged the PR into the active 0.9.6 branch.

Thank you.

> I'm curious if you can use the branch reference instead of the release tag for some time in your project? I try to bundle a few fixes or features per each release (unless there is a hotfix for a pressing issue), but so far there is just this PR, so I would ideally release later when more work is accumulated. Please let me know!

This no problem @nalexn ! A new release is not urgent; we can reference the branch or SHA for now 👍

PR #236, opened 2023-04-04
add onReceive after inspection function

Karllas:
Currently there is no way to do inspection after some action + delay which for example might be needed when you click some button and there is some processing and we want to assert something after processing is done


nalexn:
Thanks for adding this!

Issue #235, opened 2023-04-01
Viewinspector seems broken after updating Xcode to 14.3 / iOS 16.4


ztolley:
I have an existing project that has stopped working with view inspector after updating Xcode and when I try to start a new project it won't build when I add view inspector:

## New project - Steps to reproduce
* Upgrade to Xcode 14.3 with it's support for iOS 16.4
* Create a new iOS swift application with tests
* Build and run application, works fine
* Run unit test, works fine
* Go to project, package dependencies and add ViewInspector from GitHub (0.9.5)
* Rebuild project, fails.

<img width="1605" alt="Screenshot 2023-04-01 at 13 02 21" src="https://user-images.githubusercontent.com/56056/229287451-3e43ad0a-7a8e-4115-a033-1892125bc889.png">

## Existing project  (set to iOS )
* Create a simple layout with a text field that uses accessibility labels and identifiers
* Create tests to find and test the text element using search search and queries to find by identifiers. Searching by text works but identifier based tests fail. 

<img width="1633" alt="Screenshot 2023-04-01 at 13 15 20" src="https://user-images.githubusercontent.com/56056/229288227-560faf3d-fc5a-46de-88dd-a4c187119dbe.png">

Before updating to Xcode 14.3 tests to find things by the identifiers worked, I had a whole suite of tests that used that method to test the content of a thing, e.g.

```{swift}
func testTemp() throws {
    let view = TackDetail(sailDetail: getSailDetail())

    XCTAssertEqual(try
                   view.inspect()
        .find(viewWithAccessibilityIdentifier: "timeAboveMaxTemp")
        .text()
        .string()
    , "17460")
}
```

kpacholak:
Hi. I also had a problem after upgrading to Xcode 14.3
```Toggle's tap() and isOn() are currently unavailable for inspection on iOS 16. Situation may change with a minor OS version update. In the meanwhile, please add XCTSkip for iOS 16 and use an earlier OS version for testing.```

<img width="801" alt="Zrzut ekranu 2023-04-1 o 18 05 11" src="https://user-images.githubusercontent.com/63740434/229301755-4d12ab44-6038-47c2-a880-30605b4f6f31.png">

Tyler-Keith-Thompson:
I think your XCTest symbol issue is a red herring. I just tried to repro this and didn't see the issue. It's possible you accidentally added ViewInspector to your app target instead of your unit testing target. I'm trying to debug the `find` issues, though.

[...]

PR #237, opened 2023-04-05
Fix for iOS 16.4 issues with accessibility identifiers and labels

Tyler-Keith-Thompson:
Fix for #235 

Do note that there's also an issue with TImeLineView with the latest. I chose not to address it because getting out a fix for this seems far more urgent.

I didn't have to modify tests because the existing tests exposed the problem (and verified the solution)


nalexn:
Many thanks for looking into this, Tyler!

nalexn:
`TImeLineView` tests are fixed now

Tyler-Keith-Thompson:
Nice! I had a hunch you’d knock that out super quick. 

FWIW MacStadium will give you a dedicated runner that’s a physical mac (for free, since it’s an OSS project), it’s easier to keep up to date since GitHub tends to lag behind quite a bit.

anyways it should be possible to use an Xcode version Manager like Xcodes and scheduled CI jobs to catch issues like this in the future, since you test suite is so robust.

PR #238, opened 2023-04-08
handle use of FormatStyle in Text

mayoff:
This commit allows the `string()` method of `InpsectableView` to work when a `LocalizedStringKey` has an interpolation that uses a `FormatStyle`, like this:

```
let sut = Text("Test \(1234, format: .number) and \(0.5678, format: .percent)")
XCTAssertEqual(try sut.inspect().text().string(), "Test 1,234 and 56.78%")
```


nalexn:
Thank you, Rob!

PR #242, opened 2023-04-20
Added get selected value from SwiftUI.Picker component

tobywoollaston:
With TextField ViewType I am able to get the current input value with .input()

I wanted to be able get the currently selected value from a Picker ViewType without the need to inject the binding attribute or have public getters on the view.

This commit allows the `selectedValue()` method of `InpsectableView<ViewType.Picker>` to get the current value that is a binding.

```
let binding = Binding<String>(wrappedValue: "First Option")
let view = Picker(selection: binding, label: Text("Title")) {
        Text("First Option").tag(0)
        Text("Second Option").tag(1)
}
try view.inspect().picker().select(value: "Second Option")
        
XCTAssertEqual("Second Option", try view.inspect().picker().selectedValue())
```

### DIFF
--- a/Sources/ViewInspector/SwiftUI/Picker.swift
+++ b/Sources/ViewInspector/SwiftUI/Picker.swift
@@ -75,6 +75,27 @@ public extension InspectableView where View == ViewType.Picker {
         }
         casted.forEach { $0.wrappedValue = value }
     }
+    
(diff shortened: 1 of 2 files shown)


nalexn:
Thanks for the PR @tobywoollaston ! I tweaked the API a bit to align better with the other functions

PR #247, opened 2023-06-14
[WIP] Closures in environment block view inspection

bachand:
This PR is not ready to be merged. I'm finding that we can't inspect views when there is an environment value that is a closure. Boxing the closure in a struct avoids the issue. So far I do not have a fix. I've written tests to demonstrated what I described.

PR #248, opened 2023-06-14
Fix inspecting when there is a closure environment value

bachand:
## Change summary

We have noticed that applying an environment value that has a closure type to a view causes the view to become uninspectable. I've added a failing test to this PR to demonstrate this (see the test added to EnvironmentModifiersTests.swift).

After digging in, I think the cause of this is the implementation of `replacingGenericParameters(…)` in Inspector.swift. The current implementation will scan a type name string for `<` and `>` characters. When there is a generic parameter with a closure type (e.g. `SomeType<(String) -> Void>`) the scanning code gets tripped up on the `>` in the closure.

I revised the implementation of `replacingGenericParameters(…)` to only consider the first `<` and last `>`. With this change, the test that was originally failing began to pass.

## Reviewers

@nalexn when you have a chance I would love if you can take a look at this PR and see what you think!
@rafael-assis with whom I am working on using ViewInspector at our company

### DIFF
--- a/Sources/ViewInspector/Inspector.swift
+++ b/Sources/ViewInspector/Inspector.swift
@@ -116,19 +116,10 @@ private extension String {
     func replacingGenericParameters(_ replacement: String) -> String {
         guard let start = self.firstIndex(of: "<")
         else { return self }
-        var balance = 1
-        var current = self.index(after: start)
-        while balance > 0 && current < endIndex {
-            let char = self[current]
-            if char == "<" { balance += 1 }
-            if char == ">" { balance -= 1 }
-            current = self.index(after: current)
-        }
-        if balance == 0 {
-            return String(self[..<start]) + replacement +
-                String(self[current...]).replacingGenericParameters(replacement)
-        }
-        return self
+        guard let end = self.lastIndex(of: ">")
+        else { return self}
+
+        return String(self[..<start]) + replacement + String(self[index(after: end)...])
     }
 }
 
--- a/Sources/ViewInspector/Inspector.swift
+++ b/Sources/ViewInspector/Inspector.swift
@@ -116,10 +116,39 @@ private extension String {
     func replacingGenericParameters(_ replacement: String) -> String {
         guard let start = self.firstIndex(of: "<")
         else { return self }
-        guard let end = self.lastIndex(of: ">")
-        else { return self}
+        var balance = 1
+        var current = self.index(after: start)
+        while balance > 0 && current < endIndex {
+            let char = self[current]
+            if char == "<" { balance += 1 }
+            if char == ">" {
+                guard let indexOfPreviousChar = index(
+                    current,
+                    offsetBy: -1,
+                    limitedBy: startIndex)
+                else
+                {
+                    // We don't expect to get here since it means that ">" was the first character
+                    // of the string. This should not be possible with a string representing a Swift
+                    // type. Since we don't have a well-formed string, we bail out.
+                    return self
+                }
 
-        return String(self[..<start]) + replacement + String(self[index(after: end)...])
+                let previousChar = self[indexOfPreviousChar]
+                if previousChar == "-" {
+                    // We've found the "->" arrow for a closure type. Ignore this ">".
+                }
+                else {
+                    balance -= 1
+                }
+            }
+            current = self.index(after: current)
+        }
+        if balance == 0 {
+            return String(self[..<start]) + replacement +
+                String(self[current...]).replacingGenericParameters(replacement)
+        }
+        return self
     }
 }
 
--- a/Tests/ViewInspectorTests/InspectorTests.swift
+++ b/Tests/ViewInspectorTests/InspectorTests.swift
@@ -55,6 +55,8 @@ final class InspectorTests: XCTestCase {
         XCTAssertEqual(name2, "Struct1")
         let name3 = Inspector.typeName(value: Struct3<Int>(), generics: .remove)
         XCTAssertEqual(name3, "Struct3")
+        let name4 = Inspector.typeName(value: Struct3<(String) -> Void>(), generics: .remove)
+        XCTAssertEqual(name4, "Struct3")
     }
     
     func testTypeNameType() {
@@ -62,6 +64,10 @@ final class InspectorTests: XCTestCase {
         XCTAssertEqual(name1, "Struct3<Int>")
         let name2 = Inspector.typeName(type: Struct1.self)
         XCTAssertEqual(name2, "Struct1")
+        let name3 = Inspector.typeName(
+            type: ModifiedContent<EmptyView, _EnvironmentKeyWritingModifier<(String) -> Void>>.self,
+            generics: .remove)
+        XCTAssertEqual(name3, "ModifiedContent")
     }
     
     func testPrintValue() {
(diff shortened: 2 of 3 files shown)


Tests/ViewInspectorTests/InspectorTests.swift
```diff
@@ -64,6 +64,10 @@ final class InspectorTests: XCTestCase {
         XCTAssertEqual(name1, "Struct3<Int>")
         let name2 = Inspector.typeName(type: Struct1.self)
         XCTAssertEqual(name2, "Struct1")
+        let name3 = Inspector.typeName(
```
bachand, line 67: This type is similar to the type that I am having issues with in https://github.com/nalexn/ViewInspector/pull/247#discussion_r1229999348

Tests/ViewInspectorTests/InspectorTests.swift
```diff
@@ -55,13 +55,19 @@ final class InspectorTests: XCTestCase {
         XCTAssertEqual(name2, "Struct1")
         let name3 = Inspector.typeName(value: Struct3<Int>(), generics: .remove)
         XCTAssertEqual(name3, "Struct3")
+        let name4 = Inspector.typeName(value: Struct3<(String) -> Void>(), generics: .remove)
+        XCTAssertEqual(name4, "Struct3")
```
bachand, line 59: Before my changes this test failed with:

> testTypeNameValue(): XCTAssertEqual failed: ("Struct3 ()>") is not equal to ("Struct3")

Tests/ViewInspectorTests/InspectorTests.swift
```diff
@@ -55,13 +55,19 @@ final class InspectorTests: XCTestCase {
...
     
     func testTypeNameType() {
         let name1 = Inspector.typeName(type: Struct3<Int>.self)
         XCTAssertEqual(name1, "Struct3<Int>")
         let name2 = Inspector.typeName(type: Struct1.self)
         XCTAssertEqual(name2, "Struct1")
+        let name3 = Inspector.typeName(
+            type: ModifiedContent<EmptyView, _EnvironmentKeyWritingModifier<(String) -> Void>>.self,
+            generics: .remove)
+        XCTAssertEqual(name3, "ModifiedContent")
```
bachand, line 72: Before my change this test failed with:

> testTypeNameType(): XCTAssertEqual failed: ("ModifiedContent>") is not equal to ("ModifiedContent")

Sources/ViewInspector/Inspector.swift
```diff
@@ -116,19 +116,10 @@ private extension String {
...
-        var balance = 1
-        var current = self.index(after: start)
-        while balance > 0 && current < endIndex {
-            let char = self[current]
-            if char == "<" { balance += 1 }
-            if char == ">" { balance -= 1 }
-            current = self.index(after: current)
-        }
-        if balance == 0 {
-            return String(self[..<start]) + replacement +
```
bachand: The current implementation is recursive. I'm having trouble determining how the recursion would have an effect in practice.

Tests/ViewInspectorTests/ViewModifiers/EnvironmentModifiersTests.swift
```diff
@@ -17,6 +17,15 @@ final class ViewEnvironmentTests: XCTestCase {
         let sut2 = EmptyView().environment(\.colorScheme, .light)
         XCTAssertEqual(try sut2.inspect().emptyView().environment(\.colorScheme), .light)
     }
+
+    func testClosureEnvironmentValue() {
+        let value: (_ value: String) -> Bool = { value in value == "hello world" }
+        let sut = EmptyView().environment(\.testHandlerClosure, value)
+        XCTAssertNoThrow(try sut.inspect().emptyView())
+        XCTAssertNoThrow(try sut.inspect().emptyView().environment(\.testHandlerClosure))
```
bachand, line 25: Before my change this line failed with:
> testClosureEnvironmentValue: XCTAssertNoThrow failed: threw error "emptyView() found ModifiedContent<SwiftUI.EmptyView, SwiftUI._EnvironmentKeyWritingModifier<(Swift.String) -> Swift.Bool>> instead of EmptyView"

Tests/ViewInspectorTests/ViewModifiers/EnvironmentModifiersTests.swift
```diff
@@ -17,6 +17,15 @@ final class ViewEnvironmentTests: XCTestCase {
         let sut2 = EmptyView().environment(\.colorScheme, .light)
         XCTAssertEqual(try sut2.inspect().emptyView().environment(\.colorScheme), .light)
     }
+
+    func testClosureEnvironmentValue() {
+        let value: (_ value: String) -> Bool = { value in value == "hello world" }
+        let sut = EmptyView().environment(\.testHandlerClosure, value)
+        XCTAssertNoThrow(try sut.inspect().emptyView())
```
bachand, line 24: Before my change this line failed with:
> testClosureEnvironment Value (: XCTAssertNoThrow failed: threw error "empty View() found ModifiedContent<SwiftUI.EmptyView, SwiftUI._EnvironmentKeyWritingModifier<(Swift.String) -> Swift.Bool>> instead of EmptyView"

Tests/ViewInspectorTests/InspectorTests.swift
```diff
@@ -55,13 +55,19 @@ final class InspectorTests: XCTestCase {
...
+        XCTAssertEqual(name4, "Struct3")
     }
     
     func testTypeNameType() {
         let name1 = Inspector.typeName(type: Struct3<Int>.self)
         XCTAssertEqual(name1, "Struct3<Int>")
         let name2 = Inspector.typeName(type: Struct1.self)
         XCTAssertEqual(name2, "Struct1")
+        let name3 = Inspector.typeName(
+            type: ModifiedContent<EmptyView, _EnvironmentKeyWritingModifier<(String) -> Void>>.self,
```
bachand, line 70: This is similar to the type that I saw in the failure for the test that I added in EnvironmentModifiersTests.swift:
Screenshot 2023-06-14 at 3 54 59 PM

Sources/ViewInspector/Inspector.swift
bachand: I realized that the recursion could be needed for tuples. I redid the implementation to keep the recursion.

Tests/ViewInspectorTests/InspectorTests.swift
```diff
@@ -55,13 +55,21 @@ final class InspectorTests: XCTestCase {
         XCTAssertEqual(name2, "Struct1")
         let name3 = Inspector.typeName(value: Struct3<Int>(), generics: .remove)
         XCTAssertEqual(name3, "Struct3")
+        let name4 = Inspector.typeName(value: Struct3<(String) -> Void>(), generics: .remove)
+        XCTAssertEqual(name4, "Struct3")
+        let name5 = Inspector.typeName(value: (Struct3<Int>(), Struct3<String>()), generics: .remove)
+        XCTAssertEqual(name5, "(Struct3, Struct3)")
```
bachand, line 61: This test passed before my changes.

nalexn:
Thank you @bachand 🙏

PR #250, opened 2023-06-28
Fix Objective-C class EnviromentObject injection

pilot34:
Inspection of SwiftUI views with Objective-C classes injected as EnvrionmentObjects doesn't work.

When we take `String(reflecting:)` for `EnvironmentObject<SomeObjcClass>` it returns `EnvironmentObject<SomeObjcClass>` in some cases, and `EnvironmentObject<__C.SomeObjcClass>` in other cases.

That is the reason `EnvironmentInjection.inject` cannot inject the value because of the mismatch in type names.

The second problem is the check for `NSObject` when we determine if this is `environment()` modifier or `enviromentObject()`. Changed it to the check for `ObservableObject`, because any `enviromentObject` should be observable.

### DIFF
--- a/Sources/ViewInspector/Inspector.swift
+++ b/Sources/ViewInspector/Inspector.swift
@@ -110,6 +110,10 @@ private extension String {
             let end = str.index(range.upperBound, offsetBy: .init(11))
             str.replaceSubrange(range.lowerBound..<end, with: "")
         }
+
+        // For Objective-C classes String(reflecting:) sometimes adds the namespace __C, drop it too
+        str = str.replacingOccurrences(of: "__C.", with: "")
+
         return str
     }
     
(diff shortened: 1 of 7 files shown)


Sources/ViewInspector/Inspector.swift
```diff
@@ -110,6 +110,10 @@ private extension String {
             let end = str.index(range.upperBound, offsetBy: .init(11))
             str.replaceSubrange(range.lowerBound..<end, with: "")
         }
+
+        // For Objective-C classes String(reflecting:) sometimes adds the namespace __C, drop it too
+        str = str.replacingOccurrences(of: "__C.", with: "")
```
pilot34, line 115:
```
str = str.replacingOccurrences(of: "<__C.", with: "<")
```
can be a bit more accurate and also works

nalexn:
Thanks for the PR @pilot34

PR #251, opened 2023-07-11
Modify for WatchKit app support

lodstar:
From Xcode 14 onwards, watchOS apps are now generated as a single target by default. However, in projects created with a single target, using ViewInspector was not possible. So, I made some code modifications to allow the use of ViewInspector in a single target watchOS app.
 Could you please take a look and confirm if it works? Thank you.


nalexn:
All good - thanks for the PR!

PR #255, opened 2023-08-22
Add inspection support for ViewThatFits.

rafael-assis:
## Changes

Similar to the issue fixed by #228, this change aims to fix blocking inspection for Views that are nested inside a [ViewThatFits](https://developer.apple.com/documentation/swiftui/viewthatfits) View.

`ViewThatFits` has a structure similar to a `TreeView` and needs to be inspected for its `content` property for ViewInspector to find its children.

ViewThatFits_debugger

## Discussion

This is a first step that unblocks the inspection of children Views of `ViewThatFits`.

However, the [documented](https://developer.apple.com/documentation/swiftui/viewthatfits) behavior of this View states that only one of its children will be provided (rendered) based on horizontal space. Implementing inspection in a way that all children Views are returned may be problematic as all children are likely going to be similar in content, labels and accessibility traits.

This can potentially cause multiple Views returned in an inspection done with a `findAll(where:)` call.

I will investigate the possibility of hosting the `ViewThatFits` View inside a `ViewHosting` View and validate whether only one of the children view is available. That will be the scenario that is the closest to SwiftUI running on a real simulator/device.

Another suggestion we're considering is the addition of a capability that enables the test code to pick an index of the child view wanted. Alternatively we can just always default to the first child in the `ViewThatFits`.

@nalexn we'd love to have your thoughts and recommendations on this.

## Testing

I added a test to explicitly find 2 children Views inside a `ViewThatFits`, which enforces the behavior being proposed in this PR.

## Review
@bachand 
@nalexn

### DIFF
--- a/Sources/ViewInspector/SwiftUI/ViewThatFits.swift
+++ b/Sources/ViewInspector/SwiftUI/ViewThatFits.swift
@@ -1,0 +1,24 @@
+import SwiftUI
+
+@available(iOS 16.0, macOS 13.0, tvOS 16.0, *)
+internal extension ViewType {
+    struct ViewThatFits {}
+}
+
+@available(iOS 16.0, macOS 13.0, tvOS 16.0, *)
+extension ViewType.ViewThatFits: SingleViewContent {
+
+    static func child(_ content: Content) throws -> Content {
+        let view: Any = try {
+            guard let rootContent = try? Inspector.attribute(path: "_tree|content", value: content.view) else {
+                // A ViewThatFits View renders only one of its child Views based on the available horizontal space.
+                // This inspection returns all children that can potentially be picked by `ViewThatFits`.
+                // https://developer.apple.com/documentation/swiftui/viewthatfits
+                return try Inspector.attribute(path: "content", value: content.view)
+            }
+
+            return rootContent
+        }()
+        return try Inspector.unwrap(view: view, medium: content.medium)
+    }
+}
--- a/Tests/ViewInspectorTests/SwiftUI/ViewThatFitsTests.swift
+++ b/Tests/ViewInspectorTests/SwiftUI/ViewThatFitsTests.swift
@@ -1,0 +1,26 @@
+import XCTest
+import SwiftUI
+
+@available(iOS 16.0, macOS 13.0, tvOS 16.0, *)
+final class ViewThatFitsTests: XCTestCase {
+    func testAllEnclosedChildTextViews() throws {
+        let sut = TestViewThatFits()
+        let longText = try sut.inspect().find(text: "Very long text that will only get picked if there is available space.")
+        let shortText = try sut.inspect().find(text: "Short text")
+
+        XCTAssertEqual(try longText.string(), "Very long text that will only get picked if there is available space.")
+        XCTAssertEqual(try shortText.string(), "Short text")
+    }
+}
+
+// MARK: - TestViewThatFits
+
+@available(iOS 16.0, macOS 13.0, tvOS 16.0, *)
+struct TestViewThatFits: View {
+    var body: some View {
+        ViewThatFits {
+            Text("Very long text that will only get picked if there is available space.")
+            Text("Short text")
+        }
+    }
+}
--- a/Sources/ViewInspector/Inspector.swift
+++ b/Sources/ViewInspector/Inspector.swift
@@ -298,6 +298,12 @@ internal extension Inspector {
             return try ViewType.PreferenceReadingView.child(content)
         case "PopoverContent":
             return try ViewType.PopoverContent.child(content)
+        case "ViewThatFits":
+            if #available(iOS 16.0, macOS 13.0, tvOS 16.0, *) {
+                return try ViewType.ViewThatFits.child(content)
+            } else {
+                return content
+            }
         default:
             return content
         }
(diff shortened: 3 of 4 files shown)


Sources/ViewInspector/SwiftUI/ViewThatFits.swift
```diff
@@ -0,0 +1,24 @@
...
+    struct ViewThatFits {}
+}
+
+@available(iOS 16.0, macOS 13.0, tvOS 16.0, *)
+extension ViewType.ViewThatFits: SingleViewContent {
+
+    static func child(_ content: Content) throws -> Content {
+        let view: Any = try {
+            guard let rootContent = try? Inspector.attribute(path: "_tree|content", value: content.view) else {
+                // A ViewThatFits View renders only one of its child Views based on the available horizontal space.
```
rafael-assis, line 14: This comment was added to explicitly document that we currently return all children of a `ViewThatFits` View.

cc: @bachand

Tests/ViewInspectorTests/SwiftUI/ViewThatFitsTests.swift
```diff
@@ -0,0 +1,26 @@
...
+
+@available(iOS 16.0, macOS 13.0, tvOS 16.0, *)
+final class ViewThatFitsTests: XCTestCase {
+    func testAllEnclosedChildTextViews() throws {
+        let sut = TestViewThatFits()
+        let longText = try sut.inspect().find(text: "Very long text that will only get picked if there is available space.")
+        let shortText = try sut.inspect().find(text: "Short text")
+
+        XCTAssertEqual(try longText.string(), "Very long text that will only get picked if there is available space.")
+        XCTAssertEqual(try shortText.string(), "Short text")
```
rafael-assis, line 12: The test intentionally validates that both children Views are inspectable even though in a real runtime scenario only one of them would be provided.

Sources/ViewInspector/SwiftUI/ViewThatFits.swift
```diff
@@ -0,0 +1,24 @@
+import SwiftUI
+
+@available(iOS 16.0, macOS 13.0, tvOS 16.0, *)
+internal extension ViewType {
```
bachand, line 4: Should this be public? For comparison: https://github.com/nalexn/ViewInspector/blob/0f872b50ceea6c26ac6904c29ac78899c3b86b6f/Sources/ViewInspector/SwiftUI/VStack.swift#L4-L9

Sources/ViewInspector/SwiftUI/ViewThatFits.swift
bachand, line 14:
```suggestion
                // A ViewThatFits View renders only one of its child Views based on the available space.
```

bachand: The view accepts an axis on initialization.
Screenshot 2023-08-23 at 10 39 31 AM

bachand: `Inspector.attribute(path: "_tree|content", value: content.view)` is what will return all children right? If so I wonder if this comment should go above that line instead of in the `else` of the `guard` statement.

Sources/ViewInspector/SwiftUI/ViewThatFits.swift
```diff
@@ -0,0 +1,24 @@
...
+@available(iOS 16.0, macOS 13.0, tvOS 16.0, *)
+extension ViewType.ViewThatFits: SingleViewContent {
+
+    static func child(_ content: Content) throws -> Content {
+        let view: Any = try {
+            guard let rootContent = try? Inspector.attribute(path: "_tree|content", value: content.view) else {
+                // A ViewThatFits View renders only one of its child Views based on the available horizontal space.
+                // This inspection returns all children that can potentially be picked by `ViewThatFits`.
+                // https://developer.apple.com/documentation/swiftui/viewthatfits
+                return try Inspector.attribute(path: "content", value: content.view)
```
bachand, line 17: Can you help me understand this fallback? When would we expect this to be encountered?

bachand: I see we had similar code in https://github.com/nalexn/ViewInspector/pull/228 but I want to make sure it's necessary here.

Sources/ViewInspector/SwiftUI/ViewThatFits.swift
```diff
@@ -0,0 +1,24 @@
...
+            }
+
+            return rootContent
+        }()
+        return try Inspector.unwrap(view: view, medium: content.medium)
```
bachand, line 22: @nalexn can you help me better understand what it's appropriate to invoke this `unwrap(…)` method? I am trying to understand if we should always be unwrapping when extracting content from views.

Tests/ViewInspectorTests/SwiftUI/ViewThatFitsTests.swift
```diff
@@ -0,0 +1,26 @@
+import XCTest
+import SwiftUI
```
bachand, line 2:
```suggestion
import SwiftUI
@testable import ViewInspector
```

bachand (approved):
Nice! Thank you for unblocking inspection of `ViewThatFits`, which is used in some critical views within our codebase. Will be curious to see what @nalexn thinks about this change and about whether it makes sense to provide further customization options during inspection (e.g. an option to return the first child of `ViewThatFits`, instead of all children).

Sources/ViewInspector/SwiftUI/ViewThatFits.swift
nalexn: `unwrap` should be called when you get to a user-provided view that might be "wrapped" in modifiers. For example, `AnyView(x)`. This `x` can be a standard view, a custom user-defined view, or either of these but "wrapped" in a modifier, such as `x.padding(10)`. "unwrap" makes sure the root view of "x" is revealed for ease of further inspection (but modifiers are remembered as well, just stored separately).
On the other hand, when you inspect an internal child of a standard SwiftUI view, there is no need to "unwrap" them, if you can see from the code there is no generic parameter that can alter the view and turn it from `Text` into a `ModifiedContent<Text>`

nalexn:
Thanks! I'll review the PR as soon as possible. As a side note, I feel like I've already given you this tip, but instead of `po view` you can use  `po Inspector.print(view) as AnyObject`. It's identical to `po view` for primitive views, but gives much better results on complex hierarchies.

rafael-assis:
> Thanks! I'll review the PR as soon as possible. As a side note, I feel like I've already given you this tip, but instead of `po view` you can use `po Inspector.print(view) as AnyObject`. It's identical to `po view` for primitive views, but gives much better results on complex hierarchies.

Thank you @nalexn for making time to review it. We appreciate your efforts here!

Sources/ViewInspector/SwiftUI/ViewThatFits.swift
bachand: Thank you for the explanation!

PR #263, opened 2023-10-05
[iOS 17 perf degradation fix] Changing sanitizingNamespace implementation to use regular expression pattern matching.

rafael-assis:
Hi @nalexn. we have identified this iOS 17.0 issue here at Airbnb and we'd love to get your input.
I believe we have a good solution (described below). Let myself and @bachand know your thoughts on it.
Thank you! 🙏 

## Context: performance degradation on iOS 17.0

On our validation exercises to adopt Xcode 15 in the Airbnb development environment, we noticed a **significant degradation in performance** when running tests that used ViewInspector.

We identified that that degradation only happened when running against an **iOS 17.0 simulator**.
We measured the multiple sets of tests running from Xcode 15 and comparing results between 2 iPhone 14 simulators, one running iOS 16.4 the other running iOS 17.0.
Overall we could see same sets of **tests running against iOS 17.0 would take around 3x longer** compared to the runs against iOS 16.4.

Measuring execution times of the full call stack we noticed that the logic of replacing strings like ".(unknown context at $1541f4030)" from the type name in `sanitizingNamespace()` was one of the top offenders.
In large hierarchies and tests that heavily relied on `find()` and `findAll()` calls, `sanitizingNamespace()` could get called more than 30,000 times for a single run. 

In those particular scenarios, the aggregated time spent on the all executions of `sanitizingNamespace()`got close to 8x slower when running on iOS 17.0 compared to the same runs on iOS 16.4.

## Changes

We have experimented a different approach for the logic that replaces those ".(unknown context at " strings from the namespaces that uses pattern matching with the `NSRegularExpression` type.

The results on our test runs with against larger view hierarchies showed that this new implementation has similar run times on iOS 16.4 as the previous implementations. On iOS 17.0 it improved dramatically by matching the times of iOS 16.4 and some times even showing slightly better performance on these measured run times. 

## Testing

The full set of ViewInspector tests was executed before and after the fix in both iOS 16.4 and iOS 17.0.
The scenario for those tests is pretty lightweight, which made the tests run times vary a little, but no degradation was detected in either iOS versions:

| | Before | After |
| --- | --- | ----- |
| iOS 16.4 | 4.486 seconds ViewInspector_iOS16_4_before | 4.608 seconds ViewInspector_iOS16_4_after |
| iOS 17.0 | 5.336 seconds ViewInspector_iOS17_0_before | 4.929 seconds ViewInspector_iOS17_0_after |

## Review
@bachand 
@nalexn

### DIFF
--- a/Sources/ViewInspector/Inspector.swift
+++ b/Sources/ViewInspector/Inspector.swift
@@ -110,10 +110,15 @@ internal extension Inspector {
 private extension String {
     func sanitizingNamespace() -> String {
         var str = self
-        while let range = str.range(of: ".(unknown context at ") {
-            let end = str.index(range.upperBound, offsetBy: .init(11))
-            str.replaceSubrange(range.lowerBound..<end, with: "")
-        }
+
+        let pattern = "\\.\\(unknown context at ..........\\)"
+        let regex = try! NSRegularExpression(pattern: pattern)
+        let range = NSRange(location: 0, length: str.utf16.count)
+        str = regex.stringByReplacingMatches(
+          in: str,
+          options: [],
+          range: range,
+          withTemplate: "")
 
         // For Objective-C classes String(reflecting:) sometimes adds the namespace __C, drop it too
         str = str.replacingOccurrences(of: "<__C.", with: "<")


nalexn:
Many thanks @rafael-assis . Merged

rafael-assis:
> Many thanks @rafael-assis . Merged

Great! Thank you so much for merging it so quickly @nalexn! 🙏 
We'll keep experimenting in our codebase and let you know if we can find other perf issues (and fixes for them).

Sources/ViewInspector/Inspector.swift
```diff
@@ -110,10 +110,15 @@ internal extension Inspector {
 private extension String {
     func sanitizingNamespace() -> String {
         var str = self
-        while let range = str.range(of: ".(unknown context at ") {
-            let end = str.index(range.upperBound, offsetBy: .init(11))
-            str.replaceSubrange(range.lowerBound..<end, with: "")
-        }
+
+        let pattern = "\\.\\(unknown context at ..........\\)"
```
bachand, line 114: @rafael-assis I assume that we use 10 `.`s in the regex (vs the `11` offset we see in the previous code) since we know the last character is always a `)`?

Sources/ViewInspector/Inspector.swift
```diff
@@ -110,10 +110,15 @@ internal extension Inspector {
...
+        let regex = try! NSRegularExpression(pattern: pattern)
+        let range = NSRange(location: 0, length: str.utf16.count)
```
bachand, line 116: @rafael-assis can you help me understand why we chose `str.utf16.count` vs. `str.count` or `str.utf8.count`? In my experience UTF16 encoding is less common than UTF8. Additionally, in examples I've found online I'm seeing the equivalent of `str.count` ([1](https://stackoverflow.com/a/46899134), [2](https://medium.com/@dkw5877/regular-expressions-in-swift-928561ad55c8), [3](https://stackoverflow.com/a/28503676)).

Sources/ViewInspector/Inspector.swift
rafael-assis: @bachand that is correct! 👍

bachand:
Thank you @rafael-assis and @nalexn !

Sources/ViewInspector/Inspector.swift
rafael-assis: @bachand my intention is that we could get the correct length of the string (including characters that require surrogate pairs like Emojis) since we're using the `.` in the pattern that will match any kind of character.
I think this is unlikely to happen in this scenario as it only outputs the memory address, but just to be on the safe side.

I think `NSRegularExpression` [expects](https://developer.apple.com/documentation/foundation/nsregularexpression/1412446-matchesinstring?language=objc) to deal with `NSString` objects that are [encoded by default in UTF-16](https://developer.apple.com/documentation/foundation/nsstring#1666323). But I think the Swift implementation might take care of that for us.

For reference, this is the documentation/posts I found as I was writing this message:

- [Measuring the Length of a String](https://developer.apple.com/documentation/swift/string#Measuring-the-Length-of-a-String)
- StackOverflow question: https://stackoverflow.com/questions/57525620/what-is-the-correct-length-argument-to-provide-to-nsrange-for-nsregularexpressi

bachand: Thanks for the detailed response. `str.utf16.count` seems to be correct. Wow, TIL!

Screenshot 2023-10-11 at 1 14 25 PM

PR #267, opened 2023-10-16
Add support for `ignoresSafeArea()`

Arclite:
This PR adds support for the `ignoresSafeArea(_:edges:)` view modifier.


Arclite:
Let me know if there's anything to improve in this PR. I'm mostly not sure on the return type from the inspection; it behaves a bit like `.rgba()` does, but I see that that returns a struct, not a tuple. Happy to create that and put it somewhere (not sure where) if that would be preferred.

nalexn:
Thanks @Arclite !

PR #270, opened 2023-10-30
Add support for `navigationDestination(isPresented:destination:)`

andy-vanwagoner-jn:
As `NavigationLink(isPresented:destination:label:)` & similar are deprecated, add support for `.navigationDestination(isPresented:destination:)`


nalexn:
Marvelous work, thank you @andy-vanwagoner-jn

Issue #268, opened 2023-10-23
Inspect fails to find views due to `guardType(value:namespacedPrefixes:inspectionCall:)` incorrectly failing


cameroncooke:
String extension `hasPrefix(regex:wholeMatch:)` doesn't handle reserved RegEx symbols and causes false negatives when checking type prefixes. 

This is especially true when the type's description includes context information like module name, as an example:
```
(extension in MyModule):MyModule.MyComponent
```

Currently, in my project, I get the error:

```
found (extension in MyModule):MyModule.MyComponent instead of (extension in MyModule):MyModule.MyComponent
```

This in fact of course is false, the two type strings match, the issue is the injecting of the string and treating it as `regex`.

The Regex expression `(extension in MyModule):MyModule.MyComponent` does not have any matches for the literal String `(extension in MyModule):MyModule.MyComponent`. 

The reason for this is:
- The `(` and `)` are special characters used to denote the beginning and the end of a capture group.
- The `.` between the namespace and symbol name is technically also invalid but works because it will match any single character including a literal `.`.

## Reproduction

In my case, the issue was caused because my view type was created within an extension in order to namespace it, i.e:

```swift
extension SomeType {
    struct MyComponent: View {
        ....
    }
}
```

## Regex

Also while investigating this issue the code below seems a little questionable:
```swift
static var swiftUINamespaceRegex: String { "SwiftUI(|[a-zA-Z0-9]{0,2}\\))\\." }

func removingSwiftUINamespace() -> String {
    return removing(prefix: .swiftUINamespaceRegex)
        .removing(prefix: "SwiftUI.")
}

func removing(prefix: String) -> String {
    guard hasPrefix(prefix) else { return self }
    return String(suffix(count - prefix.count))
}
```

Again injecting regex into `String.hasPrefix(_:)` in the `removing(prefix:)` method via the `removingSwiftUINamespace()` method and counting the literal "prefix" characters to return a substring.

I did consider opening a pull request and working on a fix but I don't have enough knowledge around why some of this works the way it does to feel confident about changing it.

I did however create three unit tests to exercise the issues:

BaseTypesTests.swift:
```swift
    func testContainsPrefixRegex() {
        // Arrange
        let prefixes = ["MyModule.MyComponent"]

        // Assert
        XCTAssertTrue(prefixes.containsPrefixRegex(matching: "MyModule.MyComponent"))
    }

    func testContainsPrefixRegexWithContext() {
        // Arrange
[...]

PR #273, opened 2023-11-14
[Perf improvement] Implementing memoization for sanitizeNamespace and replaceGenericParameters funcs.

rafael-assis:
Hi @nalexn. Picking up were we left off on #263, we have identified another opportunity for performance improvements in ViewInspector. @bachand and I discussed it and we'd love to get your thoughts on it.
Thank you so much for your time! 🙏 

## Context: performance improvements under larger SwiftUI View hierarchies

We have profiled and traced some of our slowest tests in Instruments showed that the 2 funcs defined in the `String` extension were the top offenders in terms of running time.

- `String.replacingGenericParameters(_:)`
- `String.sanitizingNamespace()`

The combined time of all calls to both of these functions accounted for almost 50% of the entire test running time:
| Weight(s) | Weight(%) | Self Weight | Symbol Name
| --- | --- | --- | --- |
|28.26 s | 100.0%  |	0 s | UnitTestHost (16171)
8.08 s |28.5% | 8.08 s | String.replacingGenericParameters(_:)
5.80 s | 20.5%	 | 5.80 s | String.sanitizingNamespace()

Instruments_trace

It was already known that these 2 functions are called thousands of times for a single test case in our codebase.
This fact motivated an investigation on how many times these 2 funcs are called for the exact same input in order to inform a potential decision to memoize the calls. 

A histogram of the calls to `sanitizingNamespace` for each input was built from a single feature integration test run.
We found that for some types, the func `sanitizingNamespace` thousands of times for them. For example:

```
    - key : "SwiftUI.Environment<Swift.Optional<Swift.AnyHashable>>"
    - value : 3640

    - key : "SwiftUICoreUI.ScrollEventDispatch"
    - value : 144817

    - key : "Swift.Bool"
    - value : 3906

    - key : "SwiftUI.Environment<PrimitiveTypesCoreUI.AdaptiveTraitCollection>"
    - value : 19052
 
    - key : "(PrimitiveTypesCoreUI.InteractivityState) -> SwiftUI.ModifiedContent<SwiftUI._ViewModifier_Content<PrimitiveTypesCoreUI.Interactive<PrimitiveTypesCoreUI.TextStyleViewModifier>>, PrimitiveTypesCoreUI.TextStyleViewModifier>"
    - value : 17912
```

## Changes

It was then clear that caching the output of these expensive string manipulation operations would be worth it.
This is exactly what theses changes do.
It's a simple memoization which stores the computed value to be returned directly instead of recalculating them repeatedly.

## Testing

-  Ran tests at Airbnb that use ViewInspector

Local runs of the the test used to profile ViewInspector's execution time showed a **substantial improvement of running time. During the 2 Unit Tests CI runs of the changes in this PR, we noticed an **improvement of about 86% (19 seconds)**. That particular test took ~22 seconds to run before the optimizations and  ~3 seconds** after the optimizations:

Although the unit tests in ViewInspector are more lightweight compared to the tests we run at Airbnb, we could also see a good improvement of the overall running time of all tests of the library:

| Before | After |
| --- | --- |
| ViewInspector_perf_before | ViewInspector_perf_after |

## Please review:

@nalexn 
@michael-bachand

### DIFF
--- a/Sources/ViewInspector/Inspector.swift
+++ b/Sources/ViewInspector/Inspector.swift
@@ -5,8 +5,8 @@ public struct Inspector {
 
     /// Removes the "(unknown context at <memory_address>)" portion of a type name.
     /// Calls to this method are memoized and retained for the lifetime of the program.
-    /// - Parameter typeName: The raw type name. (e.g. ``SomeTypeName.(unknown context at $138b3290c).SomePropertyName``)
-    /// - Returns: The sanitized type name. (e.g. ``SomeTypeName.SomePropertyName``)
+    /// - Parameter typeName: The raw type name. (e.g. `SomeTypeName.(unknown context at $138b3290c).SomePropertyName`)
+    /// - Returns: The sanitized type name. (e.g. `SomeTypeName.SomePropertyName`)
     static func sanitizeNamespace(ofTypeName typeName: String) -> String {
         var str = typeName
 
@@ -32,9 +32,9 @@ public struct Inspector {
     /// Replaces the generic types of a given type name with a string.
     /// Calls to this method are memoized and retained for the lifetime of the program.
     /// - Parameters:
-    ///   - typeName: The original type name. (e.g. ``SomeTypeName<SomeGenericTypeParameter>``)
+    ///   - typeName: The original type name. (e.g. `SomeTypeName<SomeGenericTypeParameter>`)
     ///   - replacement: The string to replace the generic parameters with. (e.g. "<EmptyView>")
-    /// - Returns: The type name with its generic parameters replaced. (e.g. ``SomeTypeName<EmptyView>``)
+    /// - Returns: The type name with its generic parameters replaced. (e.g. `SomeTypeName<EmptyView>`)
     static func replaceGenericParameters(inTypeName typeName: String,
                                          withReplacement replacement: String) -> String {
         // Check memoized value
@@ -96,11 +96,11 @@ public struct Inspector {
     private static var sanitizedNamespacesCache: [String: String] = [:]
 
     private static let sanitizeNamespacePatterns = [
-        "\\.\\(unknown context at ..........\\)",
+        "(\\.\\(unknown context at ..........\\))",
         // This pattern may be helpful to solve issue #268.
         // It will remain disabled until it is confirmed.
         // https://github.com/nalexn/ViewInspector/issues/268
-        //"\\(extension in [a-zA-Z0-9]*\\)\\:",
+        //"(\\(extension in [a-zA-Z0-9]*\\)\\:)",
     ]
 
     private static let sanitizeNamespaceRegex = {
--- a/Sources/ViewInspector/Inspector.swift
+++ b/Sources/ViewInspector/Inspector.swift
@@ -1,18 +1,115 @@
 import SwiftUI
 
 @available(iOS 13.0, macOS 10.15, tvOS 13.0, *)
-public struct Inspector { }
+public struct Inspector {
+
+    /// Removes the "(unknown context at <memory_address>)" portion of a type name.
+    /// Calls to this method are memoized and retained for the lifetime of the program.
+    /// - Parameter typeName: The raw type name. (e.g. ``SomeTypeName.(unknown context at $138b3290c).SomePropertyName``)
+    /// - Returns: The sanitized type name. (e.g. ``SomeTypeName.SomePropertyName``)
+    static func sanitizeNamespace(ofTypeName typeName: String) -> String {
+        var str = typeName
+
+        if let sanitized = sanitizedNamespacesCache[typeName] {
+            return sanitized
+        }
+
+        let range = NSRange(location: 0, length: str.utf16.count)
+        str = sanitizeNamespaceRegex.stringByReplacingMatches(in: str,
+                                                              options: [],
+                                                              range: range,
+                                                              withTemplate: "")
+
+        // For Objective-C classes String(reflecting:) sometimes adds the namespace __C, drop it too
+        str = str.replacingOccurrences(of: "<__C.", with: "<")
+
+        Inspector.sanitizedNamespacesCache[typeName] = str
+
+        return str
+
+    }
+
+    /// Replaces the generic types of a given type name with a string.
+    /// Calls to this method are memoized and retained for the lifetime of the program.
+    /// - Parameters:
+    ///   - typeName: The original type name. (e.g. ``SomeTypeName<SomeGenericTypeParameter>``)
+    ///   - replacement: The string to replace the generic parameters with. (e.g. "<EmptyView>")
+    /// - Returns: The type name with its generic parameters replaced. (e.g. ``SomeTypeName<EmptyView>``)
+    static func replaceGenericParameters(inTypeName typeName: String,
+                                         withReplacement replacement: String) -> String {
+        // Check memoized value
+        if let typeNameDict = Inspector.replacedGenericParametersCache[typeName],
+           let cachedResult = typeNameDict[replacement] {
+            return cachedResult
+        }
+
+        // Compute value
+        let result = computeReplacementOfGenericParameters(inTypeName: typeName,
+                                                           withReplacement: replacement)
+
+        // Store memoized value
+        var typeNameDict = Inspector.replacedGenericParametersCache[typeName] ?? [:]
+        typeNameDict[replacement] = result
+        Inspector.replacedGenericParametersCache[typeName] = typeNameDict
+
+        return result
+    }
+
+    private static func computeReplacementOfGenericParameters(inTypeName typeName: String,
+                                                              withReplacement replacement: String) -> String {
+        guard let start = typeName.firstIndex(of: "<") else {
+            return typeName
+        }
+
+        var balance = 1
+        var current = typeName.index(after: start)
+        while balance > 0 && current < typeName.endIndex {
+            let char = typeName[current]
+            if char == "<" { balance += 1 }
+            if char == ">" {
+                guard let indexOfPreviousChar = typeName.index(current,
+                                                               offsetBy: -1,
+                                                               limitedBy: typeName.startIndex) else {
+                    return typeName
+                }
+                let previousChar = typeName[indexOfPreviousChar]
+                if previousChar == "-" {
+                    // We've found the "->" arrow for a closure type. Ignore this ">".
+                } else {
+                    balance -= 1
+                }
+            }
+            current = typeName.index(after: current)
+        }
+
+        if balance == 0 {
+            return String(typeName[..<start]) +
+            replacement +
+            replaceGenericParameters(inTypeName: String(typeName[current...]),
+                                     withReplacement: replacement)
+        }
+
+        return typeName
+    }
+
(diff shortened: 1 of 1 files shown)


Sources/ViewInspector/Inspector.swift
```diff
@@ -1,18 +1,115 @@
...
+public struct Inspector {
+
+    /// Removes the "(unknown context at <memory_address>)" portion of a type name.
+    /// Calls to this method are memoized and retained for the lifetime of the program.
+    /// - Parameter typeName: The raw type name. (e.g. ``SomeTypeName.(unknown context at $138b3290c).SomePropertyName``)
+    /// - Returns: The sanitized type name. (e.g. ``SomeTypeName.SomePropertyName``)
+    static func sanitizeNamespace(ofTypeName typeName: String) -> String {
+        var str = typeName
+
+        if let sanitized = sanitizedNamespacesCache[typeName] {
```
rafael-assis, line 13: I moved the exact implementation from the `String` extension to here and added the memoization calls here and in line 26.

Sources/ViewInspector/Inspector.swift
```diff
@@ -1,18 +1,115 @@
...
+                    balance -= 1
+                }
+            }
+            current = typeName.index(after: current)
+        }
+
+        if balance == 0 {
+            return String(typeName[..<start]) +
+            replacement +
+            replaceGenericParameters(inTypeName: String(typeName[current...]),
```
rafael-assis, line 88: We still call into the memoized func to potentially leverage from memoized subtypes.

Sources/ViewInspector/Inspector.swift
```diff
@@ -1,18 +1,115 @@
...
+        }
+
+        return typeName
+    }
+
+    private static var replacedGenericParametersCache: [String : [String : String]] = [:]
+    private static var sanitizedNamespacesCache: [String: String] = [:]
+
+    // swiftlint:disable force_try
+    private static let sanitizeNamespaceRegex = try! NSRegularExpression(pattern: "\\.\\(unknown context at ..........\\)")
```
rafael-assis: We thought it'd be better to keep this regex as a constant instead of recreating it every time `sanitizeNamespace` is called.

Sources/ViewInspector/Inspector.swift
```diff
@@ -310,9 +361,9 @@ internal extension Inspector {
         }
     }
     // swiftlint:enable cyclomatic_complexity
-    
+
     static func guardType(value: Any, namespacedPrefixes: [String], inspectionCall: String) throws {
-        
+
```
rafael-assis, line 376: @nalexn do you have a preference regarding the style of tabs / spaces / empty lines?
I can revert those to the original state if you prefer. Let me know.

nalexn: Nah, I like this more. [I've googled](https://stackoverflow.com/a/72840452/2923345) how to tweak the Xcode settings to not insert spaces

nalexn:
That is a substantial performance improvement, good job! Before merging this in, I suggest an additional slight refactor of `sanitizeNamespace` in [this fasion](https://github.com/nalexn/ViewInspector/issues/268#issuecomment-1806104270).
The part where it removes `"(extension in ...)"` isn't yet confirmed to be helpful, but it'd be best to make sure your final version supports adding more patterns for prefix removal

rafael-assis:
> That is a substantial performance improvement, good job! Before merging this in, I suggest an additional slight refactor of `sanitizeNamespace` in [this fasion](https://github.com/nalexn/ViewInspector/issues/268#issuecomment-1806104270). The part where it removes `"(extension in ...)"` isn't yet confirmed to be helpful, but it'd be best to make sure your final version supports adding more patterns for prefix removal

Thank you for the super quick turnaround, @nalexn! 🙏 
I have updated  `sanitizeNamespace` to support multiple patterns as you suggested.

I left the new pattern disabled for now with a comment. Let me know what you think of the approach.

Sources/ViewInspector/Inspector.swift
```diff
@@ -1,18 +1,125 @@
...
+        "\\.\\(unknown context at ..........\\)",
+        // This pattern may be helpful to solve issue #268.
+        // It will remain disabled until it is confirmed.
+        // https://github.com/nalexn/ViewInspector/issues/268
+        //"\\(extension in [a-zA-Z0-9]*\\)\\:",
+    ]
+
+    private static let sanitizeNamespaceRegex = {
+        // swiftlint:disable force_try
+        try! NSRegularExpression(pattern: sanitizeNamespacePatterns.joined(separator: "|"))
```
rafael-assis, line 108: I believe having multiple patterns joined by an [OR operator](https://developer.apple.com/documentation/foundation/nsregularexpression#1661082) (`|`) might be faster than running the `stringByReplacingMatches` for each of them. Since we're not looking to replace different patterns with different strings, this seemed like the most performant approach.

Sources/ViewInspector/Inspector.swift
```diff
@@ -1,18 +1,125 @@
...
+
+    private static var replacedGenericParametersCache: [String : [String : String]] = [:]
+    private static var sanitizedNamespacesCache: [String: String] = [:]
+
+    private static let sanitizeNamespacePatterns = [
+        "\\.\\(unknown context at ..........\\)",
+        // This pattern may be helpful to solve issue #268.
+        // It will remain disabled until it is confirmed.
+        // https://github.com/nalexn/ViewInspector/issues/268
+        //"\\(extension in [a-zA-Z0-9]*\\)\\:",
```
rafael-assis: @nalexn I left this disabled since #268 is not yet resolved with a conclusion that this additional pattern helps.
I also left a comment for context.

Sources/ViewInspector/Inspector.swift
rafael-assis: Sounds good! I like this better too!

Sources/ViewInspector/Inspector.swift
```diff
@@ -1,18 +1,125 @@
...
+            replacement +
+            replaceGenericParameters(inTypeName: String(typeName[current...]),
+                                     withReplacement: replacement)
+        }
+
+        return typeName
+    }
+
+    private static var replacedGenericParametersCache: [String : [String : String]] = [:]
+    private static var sanitizedNamespacesCache: [String: String] = [:]
```
bachand, line 96: This strategy for caching is **not** thread-safe. We think that's OK (i.e. thread safety isn't necessary) but I wanted to flag it to @nalexn out of an abundance of caution 👍

Sources/ViewInspector/Inspector.swift
```diff
@@ -1,18 +1,125 @@
 import SwiftUI
 
 @available(iOS 13.0, macOS 10.15, tvOS 13.0, *)
-public struct Inspector { }
+public struct Inspector {
+
+    /// Removes the "(unknown context at <memory_address>)" portion of a type name.
+    /// Calls to this method are memoized and retained for the lifetime of the program.
+    /// - Parameter typeName: The raw type name. (e.g. ``SomeTypeName.(unknown context at $138b3290c).SomePropertyName``)
+    /// - Returns: The sanitized type name. (e.g. ``SomeTypeName.SomePropertyName``)
```
bachand:
```suggestion
    /// - Parameter typeName: The raw type name. (e.g. `SomeTypeName.(unknown context at $138b3290c).SomePropertyName`)
    /// - Returns: The sanitized type name. (e.g. `SomeTypeName.SomePropertyName`)
```

Nit: My understanding is that the double backticks are used to create links to other types when documentation comments are displayed in Xcode. Since these are example types, I think single backticks are more appropriate.

Sources/ViewInspector/Inspector.swift
```diff
@@ -1,18 +1,125 @@
...
+        return str
+
+    }
+
+    /// Replaces the generic types of a given type name with a string.
+    /// Calls to this method are memoized and retained for the lifetime of the program.
+    /// - Parameters:
+    ///   - typeName: The original type name. (e.g. ``SomeTypeName<SomeGenericTypeParameter>``)
+    ///   - replacement: The string to replace the generic parameters with. (e.g. "<EmptyView>")
+    /// - Returns: The type name with its generic parameters replaced. (e.g. ``SomeTypeName<EmptyView>``)
```
bachand:
```suggestion
    ///   - typeName: The original type name. (e.g. `SomeTypeName<SomeGenericTypeParameter>`)
    ///   - replacement: The string to replace the generic parameters with. (e.g. "<EmptyView>")
    /// - Returns: The type name with its generic parameters replaced. (e.g. `SomeTypeName<EmptyView>`)
```

Nit: See similar comment above for rationale.

Sources/ViewInspector/Inspector.swift
bachand: Smart!

Sources/ViewInspector/Inspector.swift
bachand: This makes sense, though I wonder if we want to wrap patterns in `(` and `)` before OR'ing them together to ensure that each pattern is considered atomically?

Sources/ViewInspector/Inspector.swift
bachand: @nalexn in case it's helpful, the [Airbnb style guide has a script](https://github.com/airbnb/swift#xcode-formatting) that you can run that enables various editor settings like this one that we find helpful in our large project.

bachand (approved):
Great work @rafael-assis and thanks for the quick review @nalexn ! 🙏

Sources/ViewInspector/Inspector.swift
rafael-assis: Makes sense! I'll update to single backticks! 👍

Sources/ViewInspector/Inspector.swift
rafael-assis: Thanks for mentioning that @bachand! 
I missed this comment and did want to flag it to @nalexn as well! 👍

Sources/ViewInspector/Inspector.swift
rafael-assis: I believe the fact that we add an `|` already treats both sides of it as individual patterns to be matched. But I don't think it's an issue adding them, and it makes it more clear too, so I added them. 👍

rafael-assis:
> That is a substantial performance improvement, good job! Before merging this in, I suggest an additional slight refactor of sanitizeNamespace in https://github.com/nalexn/ViewInspector/issues/268#issuecomment-1806104270.
The part where it removes "(extension in ...)" isn't yet confirmed to be helpful, but it'd be best to make sure your final version supports adding more patterns for prefix removal

Hi @nalexn. Do you think the changes I made fully addressed your suggestions?
Let me know if you need me to make any further adjustment before you merge this PR.

Thank you!

Sources/ViewInspector/Inspector.swift
nalexn: Yeah, no need to solve a problem that may not even exist. I'm 99% sure it should be fine without synchronization, we can add it later if needed

nalexn:
Merged. Thank you @rafael-assis and @bachand !

Issue #277, opened 2023-11-20
iOS 17: Button style check 


ichikmarev:
Hey, thanks for amazing library, spotted interesting case for iOS17 `buttonStyle` check

```swift
@available(iOS 13.0, macOS 10.15, tvOS 13.0, *)
public extension InspectableView {

    func buttonStyle() throws -> Any {
        let modifier = try self.modifier({ modifier -> Bool in
            return [
                "PrimitiveButtonStyleContainerModifier",
                "ButtonStyleContainerModifier",
                "ButtonStyleModifier",
            ].contains(where: { modifier.modifierType.hasPrefix($0) })
        }, call: "buttonStyle")
        if let style = try? Inspector.attribute(path: "modifier|style|style", value: modifier) {
            return style
        }
        return try Inspector.attribute(path: "modifier|style", value: modifier)
    }
}
```

We utilize a custom ButtonStyle with an internal `style` property. However, when conducting tests on iOS 17, it fails at the `modifier|style|style` path. Perhaps we could consider updating this to directly check the iOS version at this specific point?

nalexn:
Released in v0.9.9

PR #278, opened 2023-11-20
[iOS17] Added concrete iOS version check for button style

ichikmarev:
Issues described here – https://github.com/nalexn/ViewInspector/issues/277, I believe it can help other developers 

CleanShot 2023-11-20 at 14 35 22@2x


nalexn:
Thanks for the PR!

PR #286, opened 2023-12-28
Add NavigationLink value() attribute

andy-vanwagoner-jn:
Adds a func to get the value of `NavigationLink(value:label:)`. This allows testing to assert a particular link navigates to the correct destination value.


nalexn:
Thank you @andy-vanwagoner-jn , great job as always

PR #293, opened 2024-02-08
Ported iOS16 ToggleState enum fix from SoundCloud/Axt to ViewInspector

### LINKED ISSUE #230: Toggle's tap() and isOn() are currently unavailable for inspection on iOS 16
I created this issue to keep track of this problem.

Do you have any news? Or do you know of a workaround?

sureshjoshi:
Closes #230 

Uses the solution discovered here to handle the new SwiftUI.ToggleState (instead of Bool) problem: https://github.com/soundcloud/Axt/blob/master/Sources/Axt/Native/Toggle.swift

Essentially, for iOS16+, we use a new ToggleState enum and do some unsafe pointer shenanigans to map the local enum to the SwiftUI internals enum.

Removed the forced skips from the two remaining toggle.

There could be an outstanding question of how we should handle `.mixed` - but I treat it as a `false` in this PR.


nalexn:
That's great, thank you for the PR! This should help many people

PR #313, opened 2024-06-22
Add support for inspecting `.task(id: ...)` modifier

ntsh:
Added support for inspecting [task(id:priority:_:)](https://developer.apple.com/documentation/swiftui/view/task(id:priority:_:)) modifier.

Added unit-tests to cover different types/index.

Also updated readiness documentation.


nalexn:
Thanks for the PR @ntsh !

PR #315, opened 2024-07-09
Add callOnChange(oldValue:newValue:)

andy-vanwagoner-jn:
iOS 17 deprecated the prior `onChange` with a single callback parameter in favor of 2 new ones that can take an `initial:`, and callback with no parameters, or both old & new values.


sebskuse:
Any update on this? Would be great to have this available!

nalexn:
Merged. Thanks @andy-vanwagoner-jn !

Issue #337, opened 2024-09-27
Does testing NavigationStack's destination view probably create memory leaks during testing?


greenerchen:
Hi,

I'm learning SwiftUI, Swift concurrency, and trying TDD using ViewInspector. I use the code snippet[1] to detect possible memory leaks in my unit tests, and I found that the `@ObservedObject` property couldn't be released after test teardown. `@ObservedObject` property implies it's a MainActor instance, so my tests all annotate `@MainActor` to suppress the warnings of actor-isolated messages. I'm curious if I did the test wrong and how I can fix it. Thanks!

code snippets
1. memory leaks detection
```swift
func trackForMemoryLeaks(_ instance: AnyObject, file: StaticString = #filePath, line: UInt = #line) {
        addTeardownBlock { [weak instance] in
            XCTAssertNil(instance, "Instance should have been deallocated. Potential memory leak.", file: file, line: line)
        }
    }
```
2. system under testing (SUT)
```swift
@MainActor private func makeSUT(session: SHManagedSessionProtocol = SHManagedSessionMock()) -> MatchView {
        let matcher = ShazamMatcher(session: session)
        let sut = MatchView(matcher: matcher)
        trackForMemoryLeaks(matcher)
        return sut
    }
```
3. View using `@ObservedObject`
```swift
struct MatchView: View {
    @ObservedObject var matcher: ShazamMatcher
    @State private(set) var showResult: Bool = false
    var resultViewDidAppear: ((Self) -> Void)?
    
    var body: some View {
        NavigationStack {
            VStack {
                switch matcher.state {
                case .matched:
                    ShazamStartView()
                        .accessibilityIdentifier("match_idle_state_view")
                        .onTapGesture {
                            Task {
                                try await matcher.match()
                            }
                        }
                        .onAppear {
                            showResult = matcher.state == .matched
                            resultViewDidAppear?(self)
                        }
                ...
                }
                
            }
            .navigationDestination(isPresented: $showResult, destination: {
                if let result = matcher.currentMatchResult {
                    ShazamResultView(vm: ShazamResultViewModel(result: result))
                        .onAppear(perform: {
                            matcher.reset()
                        })
                        .accessibilityIdentifier("match_matched_state_view")
                }
            })
        }
    }
}
```
4. The test
```swift
@MainActor func test_state_matched() async throws {
        var sut = makeSUT()
        
        sut.matcher.state = .matched
[...]

PR #345, opened 2024-10-18
Update guide.md for async inspection with NavigationDestination

greenerchen:
When this pr is merged, it will also close the relative issue #337 .

### DIFF
--- a/guide.md
+++ b/guide.md
@@ -626,6 +626,19 @@ try await ViewHosting.host(sut) { sut in
 }
 ```
 
+If you're writing an `async` test for `NavigationDestination`, you might need to manually call `ViewHosting.expel()` to expel the current NavigationDestination view and let the implicit expel the NagivationStack base view for reducing memory consumption and memory leaks when the view holds a @ObservedObject.
+
+```swift
+let sut = TestDisplayByStateView(state: .error)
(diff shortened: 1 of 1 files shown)


greenerchen:
Hi @nalexn, kindly review my pr. Let me know if any feedback. I realized that I forgot to assign you as reviewer two weeks ago.

nalexn:
Hey @greenerchen ,

I should thank you for raising the concern in the #337 and contributing the PR for the guide - this helped to identify a library - spread issue where I forgot to call "expell" on pretty much all tests, and this wasn't mentioned in the guide either, so others followed my example and potentially have the same issue in their tests.

I did correct all tests, as well as the guide mentioning your PR and issue number. I figured it's easier to do it separately from this PR (which I'll close) because the amendment isn't accurate - the `async` version of view hosting [doesn't need expelling](https://github.com/nalexn/ViewInspector/blob/0.10.1/Sources/ViewInspector/ViewHosting.swift#L25), it's conversely, the sync version that needs expelling. Also, the issue you found is more generic and applies not only to the case with testing the `NavigationDestination`, so mentioning it explicitly in the guide isn't needed.

Big contribution, nonetheless. Thank you again!

Issue #348, opened 2024-10-24
enhancement request for .callRefreshable()


rzagelow:
Would be awesome to have support for calling a refreshable method similar to `callOnAppear()`.. something like this:

> view.inspect().find(ViewType.VStack.self).callRefreshable()

nalexn:
Added with 0.10.1

PR #350, opened 2024-11-07
Add support for callRefreshable()

andy-vanwagoner-jn:
Adds support for `callRefreshable()`

iOS 17 changed the modifier, but both before & after got branches that worked. `readiness.md` was also updated to show support.

nalexn/ViewInspector#348


nalexn:
👍

PR #351, opened 2024-11-16
Fix async inspect() deadlock bug

nh7a:
The async version of `inspect()` can deadlock because `callbacks` might not be populated when `visit()` reads it. While this issue is rare on a fast machine, it occurs quite frequently in environments like Xcode Cloud.

For example, running a simple test like the one below 1,000 times on my laptop never passes successfully due to the deadlock.

```swift
func testAsyncViewInspect() async throws {
  let sut = TestView(flag: false)
  try await ViewHosting.host(sut) {
    try await $0.inspection.inspect { view in
      let text = try view.implicitAnyView().button().labelView().text().string()
      XCTAssertEqual(text, "false")
    }
  }
}
```
Screenshot 2024-11-16 at 10 32 12 AM


nalexn:
Thank you!

PR #366, opened 2025-02-07
Add navigationTitle (Binding) support

andy-vanwagoner-jn:
While `navigationTitle(String | Text | LocalizedStringKey)` is a bit difficult to pin down for inspection, we can get the binding from `navigationTitle(Binding<String>)` fairly simply. This PR adds `navigationTitle() -> String`, and `setNavigationTitle(String)` methods.

I would hope the same `navigationTitle() -> String` VI method eventually is extended to handle the SwiftUI `navigationTitle(String | Text | LocalizedStringKey)` cases too.

### DIFF
--- a/Sources/ViewInspector/Modifiers/NavigationBarModifiers.swift
+++ b/Sources/ViewInspector/Modifiers/NavigationBarModifiers.swift
@@ -45,7 +45,11 @@ public extension InspectableView {
 
     @available(iOS 16.0, macOS 13.0, tvOS 16.0, watchOS 9.0, *)
     func navigationTitle() throws -> String {
-        return try navigationTitleBinding().wrappedValue
+        do {
+            return try navigationTitleBinding().wrappedValue
+        } catch {
+            throw InspectionError.notSupported("navigationTitle() is only supported with a Binding<String> parameter.")
+        }
     }
 
     @available(iOS 16.0, macOS 13.0, tvOS 16.0, watchOS 9.0, *)
--- a/Sources/ViewInspector/Modifiers/NavigationBarModifiers.swift
+++ b/Sources/ViewInspector/Modifiers/NavigationBarModifiers.swift
@@ -42,6 +42,23 @@ public extension InspectableView {
             modifierName: "TransactionalPreferenceModifier<Bool, StatusBarKey>",
             path: "modifier|value", type: Bool.self, call: "statusBar(hidden:)")
     }
+
+    @available(iOS 16.0, macOS 13.0, tvOS 16.0, watchOS 9.0, *)
+    func navigationTitle() throws -> String {
+        return try navigationTitleBinding().wrappedValue
+    }
+
+    @available(iOS 16.0, macOS 13.0, tvOS 16.0, watchOS 9.0, *)
+    func setNavigationTitle(_ value: String) throws {
+        try navigationTitleBinding().wrappedValue = value
+    }
+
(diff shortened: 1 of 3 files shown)


Sources/ViewInspector/Modifiers/NavigationBarModifiers.swift
```diff
@@ -42,6 +42,23 @@ public extension InspectableView {
...
+
+    @available(iOS 16.0, macOS 13.0, tvOS 16.0, watchOS 9.0, *)
+    func navigationTitle() throws -> String {
+        return try navigationTitleBinding().wrappedValue
+    }
+
+    @available(iOS 16.0, macOS 13.0, tvOS 16.0, watchOS 9.0, *)
+    func setNavigationTitle(_ value: String) throws {
+        try navigationTitleBinding().wrappedValue = value
+    }
```
nalexn: I'm hesitant to add the setter, as this is a side effect naturally not available in the UI. For example, the library exposes `setMapRect` on `Map` view, which is a side effect the user can cause from the UI (manually adjusting the map), so I can see a use case when this is used in tests. With the navigation title, by keeping the setter unavailable, the developer would be encouraged to change it programmatically from view's state, which would contribute to their app correctness

Sources/ViewInspector/Modifiers/NavigationBarModifiers.swift
```diff
@@ -42,6 +42,23 @@ public extension InspectableView {
             modifierName: "TransactionalPreferenceModifier<Bool, StatusBarKey>",
             path: "modifier|value", type: Bool.self, call: "statusBar(hidden:)")
     }
+
+    @available(iOS 16.0, macOS 13.0, tvOS 16.0, watchOS 9.0, *)
+    func navigationTitle() throws -> String {
+        return try navigationTitleBinding().wrappedValue
```
nalexn: I would recommend when `navigationTitleBinding` throws here (when the navigationTitle is set up in any other way but Binding), we capture the error and rethrow a `InspectionError.notSupported` and explain in the error that only the variant with `Binding` is supported

nalexn (changes requested):
Thanks for the PR! A couple of minor tweaks requested

Sources/ViewInspector/Modifiers/NavigationBarModifiers.swift
andy-vanwagoner-jn: When you use a binding to set the navigation title, it can be updated from the UI. This is the functionality we are hoping to test in our app.

Screenshot 2025-02-10 at 9 59 08 AM
Screenshot 2025-02-10 at 9 59 17 AM

Sources/ViewInspector/Modifiers/NavigationBarModifiers.swift
andy-vanwagoner-jn: Sounds like a good idea. I'll try to implement that.

andy-vanwagoner-jn:
@nalexn Let me know if that's acceptable. The system does provide UI automatically for editing a bound navigation title, and I updated the error to indicate only the binding overload is currently supported.

Sources/ViewInspector/Modifiers/NavigationBarModifiers.swift
nalexn: In this case yeah, it's the user causing the state change that you otherwise cannot imitate. I would also do the same as for the getter, that is rethrow a meaningful error about API limitation

PR #369, opened 2025-04-01
Confine implicit any to compiler version 6.0

andy-vanwagoner-jn:
Xcode 16.3 with Swift 6.1 removes the implicit `AnyView`s except when actually previewing. In the long run this is simpler, though it's a bit of a pain here in a library that needs to support both way for a time.


nalexn:
Thank you @andy-vanwagoner-jn !

Issue #371, opened 2025-04-11
Xcode 16.3 `.accessibilityValue()` Throws An Error


weisunOW:
## Description

Calling `InspectableView.accessibilityValue()` throws an error below using Xcode 16.3. The same test procedure works on Xcode 16.2.

```Shell
Caught error: Description does not have 'some' attribute
```

LLDB output

```Shell
(lldb) po try sut.accessibilityValue()
▿ Description does not have 'some' attribute
  ▿ attributeNotFound : 2 elements
    - label : "some"
    - type : "Description"
```

Xcode version: 16.3
iOS SDK Version: 18.4 
Swift Version: 5.9
Test framework: Swift Testing or XCTest

## Test Failure

Failed test cases in `ViewInspector` package when running with Xcode 16.3;

* `ViewAccessibilityTests.testAccessibilityValueInspection`
* `ViewAccessibilityTests.testAccessibilityMultipleAttributes`

## Example

```Swift
  @Test
  func accessibilityValue() throws {
    let value = "abc"

    let sut = try EmptyView()
      .accessibilityValue(value)
      .inspect()
      .emptyView()

    let inspectedValue = try sut.accessibilityValue() // Error is thrown here

    #expect(try inspectedValue.string() == value)
  }
```

## Root Cause

I think the SwiftUI on iOS 18.4 have changed the descendant path of accessibility value.

See lldb output or `property.value` in `v3AccessibilityElement(...)` method.
```Shell
(lldb) po property.value
▿ Optional<AccessibilityValueStorage>
  ▿ some : AccessibilityValueStorage
    - value : nil
    ▿ description : Description
      ▿ text : 1 element
        ▿ 0 : Text
          ▿ storage : Storage
            - verbatim : "abc"
          - modifiers : 0 elements
```

## Potential Fix

Not sure if we want to refactor the existing condition in this method, but this is a quick and dirty fix to make unit tests pass.

```diff
diff --git a/Sources/ViewInspector/Modifiers/AccessibilityModifiers.swift b/Sources/ViewInspector/Modifiers/AccessibilityModifiers.swift
index 1976d33..c187afb 100644
--- a/Sources/ViewInspector/Modifiers/AccessibilityModifiers.swift
+++ b/Sources/ViewInspector/Modifiers/AccessibilityModifiers.swift
@@ -38,7 +38,17 @@ public extension InspectableView {
     func accessibilityValue() throws -> InspectableView<ViewType.Text> {
         let text: Text
[...]

PR #372, opened 2025-04-11
Fix accessibility value inspection for iOS 18.4

weisunOW:
This PR fixes the issue raised in #371 

- updated element search path for the accessibility element of `accessibilityValue`

This fix will allow the following unit tests to pass on Xcode 16.3

* `ViewAccessibilityTests.testAccessibilityValueInspection`
* `ViewAccessibilityTests.testAccessibilityMultipleAttributes`


nalexn:
Thank you @weisunOW !

PR #377, opened 2025-04-20
Add testStrokeContent to demonstrate how

dbashtani:
I couldn’t find a clear way to test shape border color, so I added this test as a helpful example. It aims to guide others facing the same challenge and serve as living documentation.


nalexn:
Thank you for the PR!

PR #381, opened 2025-05-03
Make customInspectableContent MainActor isolated

gsabran:
Looks like `customInspectableContent` can be MainActor isolated, which will play more nicely with strict swift concurrency.

Screenshot 2025-05-03 at 10 27 41 AM


nalexn:
Thank you for the PR!

PR #396, opened 2025-08-08
Supply size parameter via ViewHosting method

sebskuse:
The `(V) async throws -> Void` variant of this method passes on its `size`, the one that provides no parameter has it as a parameter, but it's not passed on.


nalexn:
Good catch!

PR #399, opened 2025-09-11
Add support for TextEditor changes in iOS 26

andy-vanwagoner-jn:
Existing `input()` and `setInput()` were broken by internal changes to `TextEditor` in iOS/macOS 26.
There were also additional properties added in iOS/macOS 26 for attributed string bindings, and text selection.

This fixes the broken methods for previous simple TextEditor cases, and adds support for the attributed input & text selection.


nalexn:
Great work Andy

Issue #395, opened 2025-08-07
[Xcode 26] Systemic failures to find subviews, "Search did not find a match"


KentLottis-jwn:
We're observing system reproducible failures across dozens of tests and view where ViewInspector just throws "Search did not find a match". I know that some of our own internal type-inspection tools are tripping over changes to the run-time, this isn't entirely surprising - it's just Apple being Apple. Just wondering: is this identified as a know issue that's getting attention as Xcode 26 gets closer to shipping?

If this is news to you, I can distill down a repro case, but figured I'd ask first.

nalexn:
Hey - yea, I'm aware of the failing tests in Xcode 26, done a brief exploration - the number of internal SwiftUI changes is much smaller than the total failing tests count, there are a few that cause multiple tests to fail. That being said, I'm very constrained on the available time, so any help would be great

PR #400, opened 2025-09-12
Fix accessibility label & value, & lazy stacks & grids in iOS 26

andy-vanwagoner-jn:
Generalizes the fix for accessibilityIdentifier to apply it to accessibilityLabel & accessibilityValue. There are more accessibility changes needed for iOS 26, but they are a bit more complicated.

Fixed HDR colors in iOS 26.

Fixed layout changes in Lazy{H|V}{Stack|Grid}.


andy-vanwagoner-jn:
I'd guess this PR would go a long way to help #395.

marcos1262:
I tested this branch locally and it fixed my issues with macOS 26 as well

nalexn:
Thank you, Andy!

PR #410, opened 2025-11-18
Fix issue #405: Add macOS exclusion guards for iOS-only SwiftUI types

### LINKED ISSUE #405: ViewInspector fails to compile on macOS SDK 26
# ViewInspector fails to compile on macOS SDK 26

## Environment
- macOS SDK Version: 26.0
- Xcode Version: 26.0.1 (Build 17A400)
- Swift Version: 6.2 (swift-6.2-RELEASE)
- ViewInspector Version: 0.9.7 - 0.10.3 (tested)
- Platform: macOS

## Issue
ViewInspector fails to compile on macOS SDK 26 because it uses SwiftUI types that are iOS-only and not available on macOS.

### VideoPlayer.swift
```
error: cannot find type 'VideoPlayer' in scope
extension VideoPlayer: SingleViewProvider {
           `- error: cannot find type 'VideoPlayer' in scope
```

### SignInWithAppleButton.swift
```
error: cannot find type 'SignInWithAppleButton' in scope
func labelType() throws -> SignInWithAppleButton.Label {
                               `- error: cannot find type 'SignInWithAppleButton' in scope
```

### Map.swift and MapAnnotation.swift
```
error: cannot find type 'MapAnnotation' in scope
error: cannot find type 'MapMarker' in scope
error: cannot find type 'MapPin' in scope
error: cannot find type 'MapUserTrackingMode' in scope
error: cannot find type 'MapInteractionModes' in scope
error: cannot find type '_DefaultAnnotatedMapContent' in scope
error: cannot find type '_MapAnnotationData' in scope
```

## Root Cause
The SwiftUI types (`VideoPlayer`, `SignInWithAppleButton`, `Map` variants) are iOS-only and not available in macOS SDK 26. ViewInspector has conditional compilation guards for `canImport(AVKit)` and `canImport(AuthenticationServices)` but does not exclude these code paths for macOS.

## Reproduction
1. Create a Swift Package targeting macOS
2. Add ViewInspector as a test dependency
3. Attempt to build on macOS SDK 26
4. Observe compilation errors

## Proposed Solution
Add macOS exclusion guards to the affected files:

- `Sources/ViewInspector/SwiftUI/VideoPlayer.swift` - Wrap in `#if !os(macOS)`
- `Sources/ViewInspector/SwiftUI/SignInWithAppleButton.swift` - Wrap in `#if !os(macOS)`  
- `Sources/ViewInspector/SwiftUI/Map.swift` - Change guard to `#if !os(macOS) && canImport(MapKit)`
- `Sources/ViewInspector/SwiftUI/MapAnnotation.swift` - Change guard to `#if !os(macOS) && canImport(MapKit)`
- `Sources/ViewInspector/ViewSearchIndex.swift` - Conditionally exclude entries in view search index

## Impact
- Blocks tests from running on macOS
- Affects CI/CD pipelines targeting macOS
- Prevents cross-platform testing workflows

## Workaround
None available without modifying ViewInspector source code. Users must either:
1. Use iOS-only test targets
2. Exclude ViewInspector from macOS test targets
3. Use alternative testing frameworks for macOS tests

## Related
These SwiftUI view types were introduced for iOS and are not available on macOS. The `canImport` checks verify frameworks are available but don't check if the SwiftUI types exist in the SDK.

schatt:
## Summary
This PR fixes compilation errors on macOS SDK 26 by adding platform exclusion guards for iOS-only SwiftUI types that are not available on macOS.

## Changes
- Wrap `VideoPlayer.swift` in `#if !os(macOS)`
- Wrap `SignInWithAppleButton.swift` in `#if !os(macOS)`
- Update `Map.swift` guard to `#if !os(macOS) && canImport(MapKit)`
- Update `MapAnnotation.swift` guard to `#if !os(macOS) && canImport(MapKit)`
- Conditionally exclude iOS-only view types (Map, SignInWithAppleButton, VideoPlayer) from `ViewSearchIndex` on macOS

## Problem
The SwiftUI types (`VideoPlayer`, `SignInWithAppleButton`, `Map` variants) are iOS-only and not available in macOS SDK 26. ViewInspector had conditional compilation guards for `canImport(AVKit)` and `canImport(AuthenticationServices)` but did not exclude these code paths for macOS.

## Solution
Added `#if !os(macOS)` guards to exclude these iOS-only types when building for macOS. The `canImport` checks verify frameworks are available but don't check if the SwiftUI types exist in the SDK.

## Testing
- ✅ Build succeeds on macOS (tested with `swift build -c release --arch arm64`)
- ✅ No linting errors
- ✅ All existing functionality preserved for iOS/tvOS/watchOS

Fixes #405

### DIFF
--- a/Sources/ViewInspector/ViewSearchIndex.swift
+++ b/Sources/ViewInspector/ViewSearchIndex.swift
@@ -6,7 +6,7 @@ import SwiftUI
 @MainActor
 internal extension ViewSearch {
     private static let index: [String: [ViewIdentity]] = {
-        let knownViewTypes: [KnownViewType.Type] = [
+        var knownViewTypes: [KnownViewType.Type] = [
             ViewType.ActionSheet.self,
             ViewType.ActionSheet.self,
             ViewType.Alert.self,
@@ -47,7 +47,6 @@ internal extension ViewSearch {
             ViewType.Link.self,
             ViewType.List.self,
             ViewType.LocationButton.self,
-            ViewType.Map.self,
             ViewType.Menu.self,
             ViewType.MenuButton.self,
             ViewType.MultiDatePicker.self,
@@ -67,7 +66,6 @@ internal extension ViewSearch {
             ViewType.ScrollViewReader.self,
             ViewType.Section.self,
             ViewType.SecureField.self,
-            ViewType.SignInWithAppleButton.self,
             ViewType.ShareLink.self,
             ViewType.Sheet.self,
             ViewType.Slider.self,
@@ -89,13 +87,19 @@ internal extension ViewSearch {
             ViewType.Toolbar.self,
             ViewType.Toolbar.Item.self,
             ViewType.Toolbar.ItemGroup.self,
-            ViewType.VideoPlayer.self,
             ViewType.ViewModifierContent.self,
             ViewType.ViewThatFits.self,
             ViewType.VSplitView.self,
             ViewType.VStack.self,
             ViewType.ZStack.self,
         ]
+        #if !os(macOS)
+        knownViewTypes.append(contentsOf: [
+            ViewType.Map.self,
+            ViewType.SignInWithAppleButton.self,
+            ViewType.VideoPlayer.self,
+        ])
+        #endif
         let identities = knownViewTypes.map { $0.viewSearchIdentity() }
         var index = [String: [ViewIdentity]](minimumCapacity: 26) // alphabet
         identities.forEach { identity in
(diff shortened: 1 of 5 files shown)


Sources/ViewInspector/ViewSearchIndex.swift
```diff
@@ -6,7 +6,7 @@ import SwiftUI
 @MainActor
 internal extension ViewSearch {
     private static let index: [String: [ViewIdentity]] = {
-        let knownViewTypes: [KnownViewType.Type] = [
+        var knownViewTypes: [KnownViewType.Type] = [
```
nalexn, line 9: That's not the best way of excluding the unavailable APIs. There is `@available(iOS ..., macOS ..., tvOS ..., watchOS ..., *)` for this purpose. I'm also not sure why we're shutting off `Map` and other views for macOS while they are actually declared as supported in the macOS SDK. There should be a more accurate fix to the issue you're experiencing, this PR is a "nuke"

PR #411, opened 2026-01-13
Safe area bar

paweldudek:
I noticed this was missing from list of supported modifiers hence this PR. 

I tried basing the implementation on existing support for `safeAreaInset` however the underlying structure is different for `safeAreaBar`, hence slightly more complex implementation.


nalexn:
Awesome work, nicely done @paweldudek

Issue #338, opened 2024-10-01
EnvironmentInjection crashes under certain conditions


shuufei:
When I run the test under certain conditions, it crashes on EnvironmentInjection in ViewInspector.

I can't identify the conditions that cause the crash because I just changed some code that I don't think is the cause of the crash and I can see the result of whether it crashes or not.

Also, the result does not change with each execution, but always crashes under certain conditions.

## Environment

- XCode: 15.4
- ViewInspector: 0.10.0
- Run Destination: Simulator iPhone 15 (17.5)

## Crash Details

EXC_BAD_ACCESS is occurring at line 47 of EnvironmentInjection.

<img width="1792" alt="スクリーンショット 2024-10-01 11 51 38" src="https://github.com/user-attachments/assets/33592c8a-5760-4dad-84db-bfca131ec714">

## Crashing Code

View under test↓

```swift

import SwiftUI

@MainActor
class ViewModel: ObservableObject {
    @Published var count: Int = 0
}

struct ContentView: View {
    @State var model: ModelA?
    
    @EnvironmentObject var viewModel: ViewModel
    
    #if DEBUG
    internal var didAppear: ((Self) -> Void)?
    #endif

    var body: some View {
        VStack {
            Text("Count: \(viewModel.count)")
            Button {
                viewModel.count += 1
            } label: {
                Text("increment")
            }
            Button {
                viewModel.count -= 1
            } label: {
                Text("decrement")
            }
        }
        .padding()
        .onAppear {
            #if DEBUG
            self.didAppear?(self)
            #endif
        }
    }
}

struct ModelA: Codable {
    let id: String
    var property1: String
    let property2: String
    var property3: String
    let property4: String
    let property5: String
    let property6: String
    var property7: [String]
    let property8: Date
    let property9: Date?
    var property10: Double?
    var property11: ModelB?
    var property12: Status

    struct ModelB: Codable, Hashable {
        let id: String
        var property1: [String]
    }

    enum Status: String, Codable {
        case ready = "READY"
        case running = "RUNNING"
    }
}

```

Test code↓

```swift
import XCTest
import ViewInspector
@testable import ViewInspectorSample

final class ViewInspectorSampleTests: XCTestCase {
    override func setUpWithError() throws {
        // Put setup code here. This method is called before the invocation of each test method in the class.
    }

    override func tearDownWithError() throws {
        // Put teardown code here. This method is called after the invocation of each test method in the class.
    }
    
    @MainActor
[...]

Issue #167, opened 2022-03-14
Feature Request: Allow tests to specify environment


Tyler-Keith-Thompson:
### The problem
I've been running into a plethora of `EnvironmentObject` issues. Seemingly, nested views that rely on it have lots of issues finding it after state changes. 

My specific use-case could be summed up as, I have a view heavily dependent on an environment object, I want ViewInspector to prove that the parent displays that view as a sheet *after* a state change. Unfortunately, before the state change in the parent, it has the environment objects, but after it keeps telling me the environment objects aren't there.

While this specific problem could be tracked down this smacks to me of ViewInspector constantly having to track down and attempt to duplicate very complex environment scenarios. 

### Potential Solution
Let consumers of the library specify the environment to ViewInspector. I believe this has several advantages:
- While consumers can use `ViewHosting.view(someView.environmentObject(injected))` this allows more explicit control with arrange-act-assert based tests. (Given my environment is <setup>, When my view <action>, Then <expected result>)
- It's possible, although I'd need to investigate further, that this would negate the need for the custom `Inspection` class to be injected into prod code. I think this is true because I believe nested views would largely be functional, therefore you could create a view wrapper that did much the same thing as `Inspection`
- While this would put the library more into "I'm going to mimic what SwiftUI does" territory, the current ViewHosting solution demonstrably does not behave as it does in production anyways, so rather than tracking down a litany of problems, this might allow us to give more stability to tests using ViewInspector. If Apple exposes a better way, or we discover one, of hosting views and *correctly* setting up the environment we can change implementation with little consumer impact.

I'll stop there because it's a lot to read. If there's interest, I'm happy to create a discussion and talk through some of my API design ideas. I am also willing to do the grunt work and submit a PR, assuming others believe this to be a worthwhile approach.

nh7a:
I'm using `ViewHosting.view(myView.environmentObject(appState))` and "async inspection", and it appears to have solved my problem on my laptop but it still 100% crashes on CI (GitHub Action & Xcode Cloud). Hopefully your idea fully solves the problem.

huwshimi:
> Cool, I'll work on a PR. FWIW my use-case seems to be more complex because distilling it down to a minimally reproducible example has been quite challenging.

I believe [this is a minimal reproduction](https://github.com/huwshimi/view-inspector-env-error) of the problem.

[...]

Issue #131, opened 2021-08-30
Question: Testing presentationMode


Tyler-Keith-Thompson:
One of the ways to programmatically dismiss a view is by using: `@Environment(\.presentationMode) var presentation` then calling `presentationMode.wrappedValue.dismiss()`. 

So if you've launched that view with a Binding of some kind, say you used `NavigationLink` with `$isActive`. When dismissed is called that binding does get set to false in production code. However I'm not sure ViewInspector triggers all the appropriate lifecycle things to have that work. 

Is there an obvious way to test this I simply don't understand? Is this untestable? Could we get it as a feature enhancement?

nalexn:
Hey, that is correct, the `@Environment(\.presentationMode)`, along with the new `@Environment(\.isPresented)` and `@Environment(\.dismiss)` are not updated nor are observed by ViewInspector because it doesn't have access to the EnvironmentValues store.

If you print a view with `@Environment(\.presentationMode)`, you'll see the variable:

```swift
 _presentationMode: Environment<Binding<PresentationMode>>
    content: Content
      keyPath: WritableKeyPath<EnvironmentValues, Binding<PresentationMode>>
        _kvcKeyPathStringPtr: Optional<UnsafePointer<Int8>> = nil
```

Unlike `@EnvironmentObject` property wrapper, which inner var references the object, this one only carries the `keyPath`, and the value is stored elsewhere, in internal SwiftUI storage.

Access to that storage is private, we cannot patch it either.

The other way could be to swizzle the `wrappedValue` getter for the `Environment` property wrapper to return a custom `Binding<PresentationMode>` that we manage - but Swift does not offer swizzling for value types.

I'll keep this ticket open in case we find a way to access the `EnvironmentValues`.

Tyler-Keith-Thompson:
I know it's random to come back to this so long later, but there might be an interesting way to access environment values: `@Environment(\.self) var env`

from: https://blog.eidinger.info/dump-swiftui-environment-efficiently

Figured I'd note that if and when anybody got around to playing with this.

mikemike396:
Anyone find a solution to use this library with `@Environment()` / latest Observation framework?

PR #416, opened 2026-03-01
[Experimental] Opt-in @Environment resolution to prevent warnings during inspection

garry-jeromson:
## Problem

When ViewInspector evaluates a custom view's `body`, any `@Environment` properties still in their `.keyPath` state trigger:

> Accessing Environment\<T\>'s value outside of being installed on a View. This will always read the default value and will not update.

In large test suites this can produce thousands of warnings (e.g. ~4,000 in ours), making it difficult to spot real issues in test output.

**Related issues:** #131, #167, #338

## Solution

This PR adds **opt-in** `@Environment` resolution that transitions fields from `.keyPath` to `.value` state before `body` evaluation, eliminating the warnings entirely.

### Enabling

```swift
// Per-test
override func setUp() {
    ViewInspectorConfig.resolveEnvironmentValues = true
}

// Or via environment variable (e.g. in your test scheme/CI)
VIEWINSPECTOR_RESOLVE_ENVIRONMENT=1
```

### Custom environment values

Test code using `.environment()` modifiers is automatically honored during resolution — both standard SwiftUI keys and custom `EnvironmentKey` types:

```swift
// Standard key override
let sut = MyView().environment(\.colorScheme, .dark)
let text = try sut.inspect().find(ViewType.Text.self).string()
XCTAssertEqual(text, "dark") // ✅ Uses the override, not the default

// Custom EnvironmentKey
let sut = MyView().environment(\.myCustomKey, "custom_value")
let text = try sut.inspect().find(ViewType.Text.self).string()
XCTAssertEqual(text, "custom_value") // ✅ Custom keys work too
```

## Technical approach

- **`EnvironmentResolvable` protocol** — conformance on `Environment<Value>` enables type-erased generic resolution via `_resolve(using:)`
- **`AnyWritableEnvironmentKeyPath` protocol** — conformance on `WritableKeyPath<EnvironmentValues, T>` enables type-erased application of `.environment()` modifier overrides
- **Raw byte manipulation** of `@frozen` `@Environment<Value>` enum layout (`.keyPath` → `.value` state transition)
- **Keypath-pointer-only matching** — when `sizeof(Value) > sizeof(KeyPath)`, Mirror-extracted copies have divergent spare payload bytes; matching only the keypath pointer (which is unique per field) avoids false negatives
- **ARC safety** — action/closure-based environment values are skipped; `buildValueStateBytes` uses raw memory allocation to avoid corrupting reference-counted keypath fields
- **Thread safety** — environment values flow through the call chain as parameters, no mutable static state

## ⚠️ Experimental — caveats

This is an **experimental** approach that relies on `@Environment`'s `@frozen` ABI-stable layout. It works reliably in our test suite but should be considered carefully before broader adoption:

1. **`@frozen` dependency** — the implementation assumes `@Environment<Value>` uses a two-case enum (`keyPath`/`value`) with a specific byte layout. This is ABI-stable but not a public API contract.

2. **Value types only** — resolution is skipped for `@Environment` values that contain reference types (closures, classes). Action-type environments like `\.dismiss` and `\.openURL` are safely ignored.

3. **Resolved values are defaults unless overridden** — without `.environment()` modifiers, values come from `EnvironmentValues()` (SwiftUI defaults). Custom keys use their `EnvironmentKey.defaultValue`.

## Files changed

| File | Description |
|------|-------------|
| `EnvironmentInjection.swift` | Core resolution logic, `ViewInspectorConfig`, `EnvironmentResolvable`, `AnyWritableEnvironmentKeyPath` |
| `ContentExtraction.swift` | Hook points: resolve before `copy.body` / `copy.body()`, thread `environmentModifiers` parameter |
| `CustomView.swift` | Pass `medium.environmentModifiers` through to `extractContent` |
| `EnvironmentInjectionTests.swift` | 16 test cases covering protocol, resolution, integration, custom keys, and regression |

## Test coverage

- ✅ Bool, enum, action environment types
- ✅ Resolution enabled/disabled (opt-in flag)
- ✅ Single and multiple `@Environment` fields
- ✅ Mixed `@Environment` + `@EnvironmentObject`
- ✅ ViewModifier with `@Environment`
- ✅ Standard key override via `.environment()` modifier
- ✅ Custom `EnvironmentKey` override via `.environment()` modifier
- ✅ Default values when no override present
- ✅ Existing `@EnvironmentObject` injection unaffected
- ✅ All 1,480 existing ViewInspector tests pass

### DIFF
--- a/Sources/ViewInspector/ContentExtraction.swift
+++ b/Sources/ViewInspector/ContentExtraction.swift
@@ -10,24 +10,30 @@ internal struct ContentExtractor {
         self.contentSource = try Self.contentSource(from: source)
     }
 
-    internal func extractContent(environmentObjects: [AnyObject]) throws -> Any {
+    internal func extractContent(environmentObjects: [AnyObject],
+                                   environmentModifiers: [EnvironmentModifier] = []) throws -> Any {
         #if swift(>=6.0)
-        return try isolatedExtractContent(environmentObjects: environmentObjects)
+        return try isolatedExtractContent(environmentObjects: environmentObjects,
+                                          environmentModifiers: environmentModifiers)
         #else
         return try MainActor.assumeIsolated {
-            try isolatedExtractContent(environmentObjects: environmentObjects)
+            try isolatedExtractContent(environmentObjects: environmentObjects,
+                                       environmentModifiers: environmentModifiers)
         }
         #endif
     }
 
     @MainActor
-    private func isolatedExtractContent(environmentObjects: [AnyObject]) throws -> Any {
+    private func isolatedExtractContent(environmentObjects: [AnyObject],
+                                        environmentModifiers: [EnvironmentModifier]) throws -> Any {
         try validateSourceBeforeExtraction()
         switch contentSource {
         case .view(let view):
-            return try view.extractContent(environmentObjects: environmentObjects)
+            return try view.extractContent(environmentObjects: environmentObjects,
+                                           environmentModifiers: environmentModifiers)
         case .viewModifier(let viewModifier):
-            return try viewModifier.extractContent(environmentObjects: environmentObjects)
+            return try viewModifier.extractContent(environmentObjects: environmentObjects,
+                                                   environmentModifiers: environmentModifiers)
         case .gesture(let gesture):
             return try gesture.extractContent(environmentObjects: environmentObjects)
         }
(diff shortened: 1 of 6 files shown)


nalexn:
This is very interesting, thanks for the PR! I hope I'll get some free time soon to explore this more and merge.

nalexn:
The PR is closed unintentionally because I'm switching branches, you can reopen the PR against new 0.10.5 or I'll do it after

PR #418, opened 2026-03-26
fix task modifiers in 26.4

andy-vanwagoner-jn:
Looks like in 26.4, the task modifier now is internally versioned 2 and takes a `sending @isolated(any)` instead of `@Sending` func.


nalexn:
Thank you Andy!

PR #421, opened 2026-06-13
Support iOS 27 introspection layout changes

nh7a:
## Summary

iOS 27 (betas 1 through 4) reshaped several SwiftUI internals that ViewInspector reflects into, causing a crash and 25 test failures. This PR restores support. All paths are version-gated (or runtime layout-detected) and keep iOS ≤26 behavior byte-identical.

| Area | iOS 27 change | Fix |
|---|---|---|
| **GeometryReader** | `GeometryProxy` layout size changed; the fixed 48/52-byte `unsafeBitCast` traps on size mismatch (**crash**) | size-agnostic zero-filled allocation |
| **Accessibility** (10 tests) | `AccessibilityProperties` named members → generic `storage: [Entry]` | new `v5` path aggregating entries across attachment modifiers, matched by value sub-path/type |
| **GlassEffect** (12 tests) | `GlassEffectModifier` wrapped in `StaticIf` (config under `trueBody`); `glass` params → keyed dict; spacing → `Smoothness` struct; transition `kind` gained an associated value | path + dict walk + `hasPrefix` |
| **StatusBar** | `statusBar(hidden:)` → `ToolbarAppearanceModifier` carrying a `Visibility` | match by `statusBar` placement role |
| **Toolbar** | lone implicit item stored directly under `content` | `content` path fallback |
| **AsyncImage** | `url` → optional `AsyncImageSource` | new path fallback |

## Verification

- **iOS 27.0 beta 4** (sim `24A5390f`, Xcode 27.0 `27A5228h`) — the 8 affected suites: **96 tests, 0 failures**. Layouts unchanged across betas 1–4.
- **iOS 27.0 beta 3** (sim `24A5380g`, Xcode 27.0 `27A5218g`) — same 8 suites: **96 tests, 0 failures**.
- **iOS 27.0 beta 2** (sim `24A5370g`, Xcode 27.0 `27A5209h`) — same 8 suites: **96 tests, 0 failures**.
- **iOS 27.0 beta 1** (sim `24A5355p`) — full suite: **1510 tests, 0 failures** (was 1 crash + 25 failures).
- **iOS 26.5** — every changed area re-verified green; no regression.

## Patterns

All changes follow existing repo precedent: version-gate ladders (newest-first, additive — matching the accessibility `v2/v3/v4` ladder), the `iOS 27 / macOS 27 / tvOS 27 / watchOS 27` platform tuple (mirroring the iOS 26 set), raw `Mirror` for structural traversal (as in `SafeAreaBar`, `Tab`), `StaticIf.trueBody` access (as in `ConditionalContent`), and runtime layout-size detection for `GeometryProxy` (the original already branched on `MemoryLayout` size).

## Notes (out of scope for this PR)

- **Deployment target under Xcode 27.** Xcode 27 raises the minimum iOS deployment target to **15.0**, but the project declares **12.0**, so a build under Xcode 27 fails with `IPHONEOS_DEPLOYMENT_TARGET is set to 12.0, but the range of supported deployment target versions is 15.0 to 27.0.x`. Verification used an `IPHONEOS_DEPLOYMENT_TARGET=15.0` build override. Raising the package's minimum platform will be needed before Xcode 27 GA — left for a separate change.

## Caveat

These are **iOS 27 beta** internal layouts (verified through beta 4). SwiftUI internals are private implementation detail, not ABI-stable, and can churn between betas and GA. Re-verify against later betas and the GA release; the version *gates* are stable but the *path strings* inside each branch may need updating.

### DIFF
--- a/Sources/ViewInspector/Modifiers/AccessibilityModifiers.swift
+++ b/Sources/ViewInspector/Modifiers/AccessibilityModifiers.swift
@@ -15,7 +15,9 @@ public extension InspectableView {
         }
         let text: Text
         let call = "accessibilityLabel"
-        if #available(iOS 26.0, macOS 26.0, tvOS 26.0, watchOS 26.0, *) {
+        if #available(iOS 27.0, macOS 27.0, tvOS 27.0, watchOS 27.0, *) {
+            text = try v5AccessibilityFirstText(path: "some|texts", call: call)
+        } else if #available(iOS 26.0, macOS 26.0, tvOS 26.0, watchOS 26.0, *) {
             text = try v4AccessibilityPropertyFirst(path: "label|some|texts", call: call)
         } else if #available(iOS 16.0, macOS 13.0, tvOS 16.0, watchOS 9.0, *) {
             if let firstText = try? v3AccessibilityElement(
@@ -40,7 +42,9 @@ public extension InspectableView {
     func accessibilityValue() throws -> InspectableView<ViewType.Text> {
         let text: Text
         let call = "accessibilityValue"
-        if #available(iOS 26.0, macOS 26.0, tvOS 26.0, watchOS 26.0, *) {
+        if #available(iOS 27.0, macOS 27.0, tvOS 27.0, watchOS 27.0, *) {
+            text = try v5AccessibilityFirstText(path: "some|description|text", call: call)
+        } else if #available(iOS 26.0, macOS 26.0, tvOS 26.0, watchOS 26.0, *) {
             text = try v4AccessibilityPropertyFirst(path: "value|some|description|text", call: call)
         } else if #available(iOS 18.4, macOS 15.4, tvOS 18.4, watchOS 11.4, *) {
             if let firstText = try v3AccessibilityElement(
@@ -74,7 +78,9 @@ public extension InspectableView {
     func accessibilityHint() throws -> InspectableView<ViewType.Text> {
         let text: Text
         let call = "accessibilityHint"
-        if #available(iOS 26.0, macOS 26.0, tvOS 26.0, watchOS 26.0, *),
+        if #available(iOS 27.0, macOS 27.0, tvOS 27.0, watchOS 27.0, *) {
+            text = try v5AccessibilityFirstText(path: nil, call: call)
+        } else if #available(iOS 26.0, macOS 26.0, tvOS 26.0, watchOS 26.0, *),
            let firstText = try v4AccessibilityProperty(
             path: "typedValue", type: [Text].self,
             call: call, { $0.accessibilityHint("") }).first {
@@ -96,7 +102,10 @@ public extension InspectableView {
     
     func accessibilityHidden() throws -> Bool {
         let call = "accessibilityHidden"
-        if #available(iOS 26.0, macOS 26.0, tvOS 26.0, watchOS 26.0, *) {
+        if #available(iOS 27.0, macOS 27.0, tvOS 27.0, watchOS 27.0, *) {
+            let raw = try v5AccessibilityElement(path: "value|rawValue", type: UInt32.self, call: call)
+            return raw != 0
+        } else if #available(iOS 26.0, macOS 26.0, tvOS 26.0, watchOS 26.0, *) {
             let value = try v4AccessibilityProperty(path: "visibility|some|value|rawValue", type: UInt32.self, call: call)
             return value != 0
         } else if #available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *) {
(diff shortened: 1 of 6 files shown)


nh7a:
This PR fixes #422.

Zeropointer11:
any progress on this? i try to prepare my project for iOS 27 and this fix would help me not to disable all my tests

Joseph-Cursio:
Independent verification on **macOS 27**, which I don't think this PR has been exercised against — the write-up covers iOS 27 betas 1–4.

Short version: it fixes the crash on macOS too, and clears more suites than claimed for iOS.

### Environment

- macOS 27.0 beta, build `26A5388g` (Darwin 27.0.0), arm64
- Xcode 26.6 (`17F113`), macOS SDK 26.5 — note this is **not** Xcode 27
- `swift test` against the package

### Results

| | `0.10.3` | this PR |
|---|---|---|
| Fatal errors | 1 | **0** |
| `GeometryReaderTests` | crashes the process | **4/4 pass** |
| Tests executed | 1504 (only by skipping `GeometryReaderTests`) | **1555** |
| Unexpected failures | 27 | **11** |

The `0.10.3` baseline could not complete a run at all — `GeometryReaderTests.testEnclosedView` traps with `Can't unsafeBitCast between types of different sizes` and kills the process, so that column required `--skip GeometryReaderTests`.

Suites failing on `0.10.3` and green with this PR on macOS: `ViewAccessibilityTests` (6), `ViewEventsTests` (4), `ViewAccessibilityActionTests` (2), `ToolbarTests` (1), `AsyncImageTests` (1).

### A second data point for the size-agnostic allocation

`MemoryLayout<GeometryProxy>` measured on two machines with an **identical toolchain** — Xcode 26.6 (`17F113`), macOS SDK 26.5, arm64 — so the OS is the only variable:

| macOS | size / stride / alignment | `0.10.3` |
|---|---|---|
| 26.5.2 (`25F84`) | **52** / 52 / 4 | works — `Allocator52` matches |
| 27.0 beta (`26A5388g`) | **76** / 76 / 4 | traps |

Plus **68** on the iOS 27 simulator per #422. Three platforms, three sizes, and the two new ones are neither 48 nor 52 — which seems like strong support for allocating dynamically here rather than extending the fixed-size ladder.

### Still failing on macOS 27 with this PR

Two clusters. I'd treat them differently:

**1. LazyVStack / LazyHStack — 10 tests, likely a real gap**

```
"LazyVStack<Text> does not have 'tree' attribute"
"LazyVStack<ForEach<ClosedRange<Int>, Int, Text>> does not have 'tree' attribute"
```

`testAlignmentInspection`, `testContentViewInspection`, `testPinnedViewsInspection`, `testSearch`, `testSpacingInspection` — the same five in each suite. This looks like the same class of internal-layout change this PR addresses elsewhere, just not covered.

**2. Localization — 12 tests, cause unconfirmed**

`TextTests.testLocalizationGlobalDefault`, `testResourceLocalizationStringNoParams`, `testLocalizedStringResourceStringWithLocaleNoParams`, plus the `ViewSearchTests` find-localized-text cases:

```
("Test") is not equal to ("Test_en")
("Test") is not equal to ("Тест_ru")
```

The base string is coming back unlocalized. **I can't tell whether this is macOS 27 or just test-bundle localization not resolving under `swift test` on macOS**, so I'd not attribute these to the OS without confirmation. The remaining `ViewSearchTests` failures (`testFindAll`, `testFindText`, `testFindButton`, `testFindCustomView`, `testFindViewWithId`, `testFindViewWithTag`, `testSkipFound`) may be downstream of the same thing, since several assert on localized strings.

### Why I ran this

A `GeometryReader` anywhere in an inspected view tree takes down the whole test process, and because the reported crash site is the Swift stdlib's `unsafeBitCast`, the visible symptom is ~115 unrelated suites failing at 0.000s with a different set each run. Took a while to trace back here. Happy to re-run against any changes, or to dig into the LazyStack `tree` attribute if that's useful.

jqsilver:
Independent verification of the `GeometryReader` fix here, against a large private iOS codebase rather than ViewInspector's own suite.

**Environment:** Xcode 27.0 beta 4 (`27A5228h`), simulator iOS 27.0 (`24A5390f`) — the same pairing listed above.

We hit the `GeometryProxy` trap from #422 in CI. Here's our stack trace

```
libswiftCore.dylib   _assertionFailure(_:_:file:line:flags:)
<TestBundle>         GeometryProxy.init()
<TestBundle>         GeometryReader.view()
<TestBundle>         ViewSearch.Traversal.search(in:condition:stopOnFoundMatch:...)
<TestBundle>         InspectableView.find(textWhere:locale:)
```

Applying **only** the `GeometryReader.swift` hunk from this PR to our vendored copy:

- **iOS 27.0 beta 4** — zero traps; test processes that previously died on their first integration test now run to completion.
- **iOS 26.5 (Xcode 26.5)** — same suite **32 tests, 0 failures**, confirming the dynamically-sized zero-fill is behavior-neutral on iOS ≤26, which was the property I most wanted to check before shipping it internally.

If there's anything I can do to help this get merged, please let me know!

nalexn:
Merged, thanks to everyone involved! This should be released after the stable Xcode 27 build is out; while it's in beta, you can pin to the branch `0.10.4`

Issue #422, opened 2026-06-15
iOS 27 Beta: unsafeBitCast fails for GeometryProxy


ciaranmul:
# Issue Summary

When running against iOS 27 simulator tests fail with:

> Swift runtime failure: Can't unsafeBitCast between types of different sizes

```swift
private extension GeometryProxy {
    struct Allocator48 {
        let data: (Int64, Int64, Int64, Int64, Int64, Int64) = (0, 0, 0, 0, 0, 0)
    }
    struct Allocator52 {
        let data: (Allocator48, Int32) = (.init(), 0)
    }
    
    init() {
        if MemoryLayout<GeometryProxy>.size == 52 {
            self = unsafeBitCast(Allocator52(), to: GeometryProxy.self)
            return
        }
        self = unsafeBitCast(Allocator48(), to: GeometryProxy.self) // <-- here
    }
}
```

The size of GeometryProxy appears to have changed - now 68 bytes, so fails when trying to cast with `Allocator48`.

# Reproduction Steps

Using Xcode 27, targeting iOS 27 simulator:

```swift
import XCTest
import ViewInspector
import SwiftUI

struct ReproductionView: View {
    var body: some View {
        GeometryReader { proxy in
            Image(systemName: "person")
        }
    }
}

class ReproductionTests: XCTestCase {
    func testReproduction() throws {
        let view = ReproductionView()
        if let image = try view.inspect().findAll(ViewType.Image.self).first {
            let path = "view(ReproductionView.self).geometryReader(0).descendants(_:)"
            XCTAssertEqual(image.pathToRoot, path)
        }
    }
}
```

nalexn:
This is pending release; The fix should be released after the stable Xcode 27 build is out; while it's in beta, you can pin to the branch `0.10.4`

scottasoutherland:
Xcode 27 GA is out. I still see this crash. And it is not 68 bytes for me it seems: 

```
(lldb) po MemoryLayout<GeometryProxy>.size
76
```

PR #424, opened 2026-08-07
Support iOS 27 LazyVStack and LazyHStack introspection

jqsilver:
## The problem

On iOS 27, `find(...)` cannot reach any view inside a `LazyVStack` or `LazyHStack`. This already fails **10 tests in the existing suites** on the iOS 27 simulator — `testSearch`, `testContentViewInspection`, `testAlignmentInspection`, `testSpacingInspection` and `testPinnedViewsInspection`, in both `LazyVStackTests` and `LazyHStackTests`.

iOS 27 flattened both types. The `tree` wrapper is gone and the content sits directly on the view:

```
LazyVStack (iOS 27):  .alignment  .spacing  .pinnedViews  .content
                      tree|content THREW: LazyVStack<Text> does not have 'tree' attribute

LazyVGrid  (iOS 27):  .tree = ResettableLazyLayoutRoot -> tree.content = Tree   (unchanged)
```

`LazyVGrid` and `LazyHGrid` delegate to the same `LazyHStack.children(_:)` but kept the earlier layout, which is why only the stacks broke — `LazyVGridTests` and `LazyHGridTests` pass on iOS 27 today.

## The fix

`children(_:)` attempts `tree|content` first and falls back to the flat `content`, so every shape that works today takes the identical path. The three custom attributes get the same treatment: when there is no `tree` they read `alignment` / `spacing` / `pinnedViews` directly off the view, composing with the existing `tree|content|root` fallback rather than replacing it.

## Verification

Existing `LazyVStackTests`, `LazyHStackTests`, `LazyVGridTests` and `LazyHGridTests`:

| | iOS 27.0 sim (`24A5390f`, Xcode 27.0 beta 4) | iOS 26.5 (Xcode 26.5) |
|---|---|---|
| before | 34 tests, **10 failures** | 34 tests, 0 failures |
| after | 34 tests, **0 failures** | 34 tests, 0 failures |

No new tests: the existing suites already cover both `children(_:)` and all three attributes, and they exercise whichever internal shape the running OS provides, so they now pass on both.

Also validated in a large private codebase, where this silently hid the entire content subtree of every scrolling screen — traversal reached 104 views on iOS 27 versus 270 on iOS 26.5, and found zero `Text` anywhere.

One note: `IPHONEOS_DEPLOYMENT_TARGET` is 12.0 in `ViewInspector.xcodeproj`, below Xcode 27's floor of 15.0, so running the suite under Xcode 27 needs `IPHONEOS_DEPLOYMENT_TARGET=15.0` on the command line. I left the project alone since raising it is your call, but it may be worth doing so iOS 27 is testable out of the box.

Related: #422 reports the `GeometryProxy` crash, which is the same family of moved internals, and #421 fixes it.

### DIFF
--- a/Sources/ViewInspector/SwiftUI/LazyHStack.swift
+++ b/Sources/ViewInspector/SwiftUI/LazyHStack.swift
@@ -34,7 +34,15 @@ public extension InspectableView where View: MultipleViewContent {
 extension ViewType.LazyHStack: MultipleViewContent {
 
     public static func children(_ content: Content) throws -> LazyGroup<Content> {
-        let view = try Inspector.attribute(path: "tree|content", value: content.view)
+        // iOS 27 flattened LazyVStack and LazyHStack: the content sits directly on the view,
+        // and there is no longer a `tree` wrapper. LazyVGrid and LazyHGrid delegate here and
+        // kept the earlier layout, so `tree|content` is attempted first.
+        let view: Any
+        if let tree = try? Inspector.attribute(path: "tree|content", value: content.view) {
+            view = tree
+        } else {
+            view = try Inspector.attribute(label: "content", value: content.view)
+        }
         return try Inspector.viewsInContainer(view: view, medium: content.medium)
     }
 }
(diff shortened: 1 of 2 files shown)


nalexn:
Thank you for the PR!

PR #426, opened 2026-08-13
Support @Observable objects in @Environment

### LINKED ISSUE #423: Observable Macro with Environment Crashes the Test
I have the following code:

```
struct LoginScreen: View {
    
    @Environment(Router.self) private var router
    
    var body: some View {
        Button("Login") {
            
            router.navigate(to: .student(.dashboard))
        }
    }
}
```

I am writing a basic test using ViewInspector.

```
 @Test
    @MainActor
    func loginButton_navigatesToStudentDashboard() async throws {
        // Given
        let router = Router()
        let sut = LoginScreen().environment(router)
        
        let button = try sut.inspect().find(button: "Login")
        try button.tap()
        
        // Then: Verify route was updated
        #expect(router.routes.first == .student(.dashboard))
    }
```

When I run the test I get the error:

SwiftUICore/Environment+Objects.swift:34: Fatal error: No Observable object of type Router found. A View.environmentObject(_:) for Router may be missing as an ancestor of this view.

Is Observable not yet supported by ViewInspector?

nh7a:
Fixes #423

## Problem

A view that takes an `@Observable` object from the environment crashes during inspection:

```swift
@Observable class Router { var routes: [Route] = [] }

struct LoginScreen: View {
    @Environment(Router.self) private var router
    var body: some View {
        Button("Login") { router.navigate(to: .dashboard) }
    }
}

let sut = LoginScreen().environment(Router())
try sut.inspect().find(button: "Login") // 💥
```

```
SwiftUICore/Environment+Objects.swift:34: Fatal error: No Observable object of type Router
found. A View.environmentObject(_:) for Router may be missing as an ancestor of this view.
```

Two causes:

1. `Content.unwrappedModifiedContent()` collected the object supplied with `.environment(_:)` only when it was an `ObservableObject`. Types using the `@Observable` macro conform to `Observable` instead, so the object never reached `Medium.environmentObjects`.
2. Even when collected, `EnvironmentInjection` only knew how to inject `@EnvironmentObject`. `@Environment(Type.self)` is a different property wrapper: `SwiftUI.Environment<T>` stores either a `KeyPath` into `EnvironmentValues` or the already resolved value. Reading the unresolved `KeyPath` outside a hosted view hierarchy is what makes SwiftUI trap above.

## Solution

Inject such objects by replacing the `keyPath` case of the property wrapper's storage with the `value` case, in the same spirit as the existing `@EnvironmentObject` forgery.

The properties to inject are found by comparing their key path with the ones SwiftUI uses for the plain and the optional `@Environment(Type.self)` declarations, instead of matching on type names. This keeps an `@Environment(\.someKey)` value of the same class type untouched.

Supported for iOS 17 / macOS 14 / tvOS 17 / watchOS 10 and above; earlier OS versions are unaffected.

## Notes

- Both `@Environment(Router.self) var router` and `@Environment(Router.self) var router: Router?` are handled, in views and in custom `ViewModifier`s.
- The existing `@EnvironmentObject` injection is unchanged: the original function body was only renamed to `injectEnvironmentObject(_:into:)`, and the new code returns early for objects that are not `Observable`.
- When a hosted view mutates an `Observable` object taken from the environment, SwiftUI drops those objects from the environment on the following re-render, so only one inspection per hosting is possible in that case. This reproduces without ViewInspector taking part in the render, so it is left as is and noted in the tests.
- Forgetting `.environment(object)` still results in SwiftUI's own fatal error rather than an `InspectionError`: object-keyed `Environment` properties cannot be told apart from `@Environment(\.someKey)` ones without knowing the object type upfront.

## Testing

New `EnvironmentObservableInjectionTests` with 10 tests, mirroring `EnvironmentObjectInjectionTests`: memory layout and forgery assumptions, direct injection, synchronous inspection with a tap, the optional declaration, `actualView()`, `didAppear` and async inspection, non-interference with an `@Environment` value of the same type, and retain/release of the injected object.

The full suite was run with `xcodebuild test` on macOS before and after the change: 1548 test cases passed before, 1558 after, no failures in either run, and no pre-existing test changed its result.

### DIFF
--- a/Sources/ViewInspector/EnvironmentInjection.swift
+++ b/Sources/ViewInspector/EnvironmentInjection.swift
@@ -1,4 +1,5 @@
 import SwiftUI
+import Observation
 
 // MARK: - EnvironmentObject injection
 
@@ -20,6 +21,11 @@ internal enum EnvironmentInjection {
         }
     }
     static func inject<T>(environmentObject: AnyObject, into entity: T) -> T {
+        let entity = injectEnvironmentObject(environmentObject, into: entity)
+        return injectEnvironment(object: environmentObject, into: entity)
+    }
+
+    private static func injectEnvironmentObject<T>(_ environmentObject: AnyObject, into entity: T) -> T {
         let type = "SwiftUI.EnvironmentObject<\(Inspector.typeName(value: environmentObject, namespaced: true))>"
         let mirror = Mirror(reflecting: entity)
         guard let label = mirror.children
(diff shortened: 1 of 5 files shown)


nalexn:
That's great, thanks @nh7a ! We should explore using this approach to add support for arbitrary type injection for `@Environment`, not just `@Observable`

PR #427, opened 2026-08-14
Support arbitrary type injection for @Environment

nh7a:
Follow-up to #426, per the suggestion to explore using that approach for arbitrary type injection, not just `@Observable`.

## What this does

`.environment(keyPath, value)` was collected in the inspection medium for reading back with `try view.environment(keyPath)`, but the value never reached the `@Environment(keyPath)` properties of the views underneath — their `body` kept reading the `EnvironmentKey`'s default value. Now it does, for any value type, without hosting the view:

```swift
struct ProfileScreen: View {

    @Environment(\.isPremiumUser) private var isPremiumUser

    var body: some View {
        if isPremiumUser { Text("Premium") } else { Button("Upgrade") { ... } }
    }
}

let sut = ProfileScreen().environment(\.isPremiumUser, true)
XCTAssertNoThrow(try sut.inspect().find(text: "Premium"))
```

The innermost modifier wins, as in the running app. Properties not covered by a modifier keep reading the key's default value.

## How

Same trick as #426 — replace the property wrapper's `keyPath` case with the `value` case — with two generalizations:

- **Layout.** #426 assumed a pointer-sized payload plus the case discriminator, which only holds for a class. `Environment<T>` is `max(8, sizeof(T)) + 1` bytes, so the layout now comes from a replica of the private `SwiftUI.Environment.Content` enum, validated against the real one before every injection. A mismatch skips the injection rather than corrupting the view, so a future SwiftUI layout change degrades to a no-op.
- **ARC.** Byte-copying a forged wrapper under-retains a multi-word payload such as `String`, freeing it prematurely. The value is now moved into the property: the destination is deinitialized (releasing the replaced key path) and receives its own strong reference.

With that in place, the `Observable` injection is a special case of the generic one — it just passes the key paths SwiftUI assigns to `@Environment(Type.self)` — so its dedicated code path from #426 is deleted. Net effect is ~110 lines on top of the object-only case, with one mechanism instead of two.

## Notes for review

- **This changes behavior of existing tests, not just adds capability.** A custom view branching on an `@Environment` value now sees ancestor `.environment(...)` values, including ones set by system modifiers that lower to `_EnvironmentKeyWritingModifier` (`.font`, `.foregroundStyle`, …). Tests asserting the old default-value behavior can flip. One in-repo test did: `testEnvironmentValueOfTheSameTypeIsUnaffected` became `testEnvironmentValueOfTheSameTypeIsInjectedSeparately`, since `@Environment(\.key)` and `@Environment(Type.self)` for the same class type are now each injected from their own modifier.
- **Known limitation, deliberate.** Properties are located by the byte pattern of the key path reference and the case discriminator, and the injection is skipped unless the match count equals the number of properties declaring that key path. A view that stores another view reading the same key path therefore keeps the default value in its own property (the stored view still gets injected when the inspection reaches it). Relaxing the count check would fix that but reopens the case it guards: a view storing a bare `KeyPath<EnvironmentValues, V>` field, where a false positive would write 9+ bytes over an 8-byte field. Chose never-corrupt over never-miss.
- **`transformEnvironment` stays unsupported.** It provides a transform closure instead of a value, which fails the type check, and it cannot shadow a value set for the same key path. A natural follow-up is resolving a value by seeding from `EnvironmentValues()[keyPath:]` and folding all matching modifiers in order, which would cover transforms too.
- `extractContent(environmentObjects:)` on `View` and `ViewModifier` gained a defaulted `environmentValues:` parameter — source-compatible for existing callers.
- `readiness.md`: `environment<V>(WritableKeyPath, V)` flipped :technologist: → :white_check_mark:.

## Testing

16 new tests in `EnvironmentValueInjectionTests` covering `Int`, `String`, `Bool`, `Optional`, a multi-field struct, a class, a system key (`\.layoutDirection`), custom `ViewModifier` properties, `actualView()`, nested custom views, duplicate key paths, innermost-wins, ARC balance in both directions, and `transformEnvironment` neither injecting nor shadowing.

Full suite is unchanged against the base: 92 pre-existing failures, all in `ViewSearchTests`/`TextTests` from test-bundle localization lookups under `swift test`. SwiftLint clean. Also verified the replica enum compiles at the package's iOS 12 deployment target.

### DIFF
--- a/Sources/ViewInspector/BaseTypes.swift
+++ b/Sources/ViewInspector/BaseTypes.swift
@@ -171,6 +171,20 @@ internal extension Content {
         let environmentModifiers: [EnvironmentModifier]
         let environmentObjects: [AnyObject]
         
+        /// The values of the enclosing `.environment(keyPath, value)` modifiers,
+        /// in the order of their application, for injecting in the `@Environment` properties.
+        #if swift(>=6.0)
+        @MainActor
+        #endif
+        var environmentValues: [EnvironmentValueInjection] {
+            return environmentModifiers.compactMap { modifier in
+                guard let keyPath = (try? modifier.keyPath()) as? AnyKeyPath,
+                      let value = try? modifier.value()
+                else { return nil }
+                return (keyPath, value)
+            }
+        }
+
         static var empty: Medium {
             return .init(viewModifiers: [],
                          transitiveViewModifiers: [],
--- a/Sources/ViewInspector/ContentExtraction.swift
+++ b/Sources/ViewInspector/ContentExtraction.swift
@@ -10,24 +10,28 @@ internal struct ContentExtractor {
         self.contentSource = try Self.contentSource(from: source)
     }
 
-    internal func extractContent(environmentObjects: [AnyObject]) throws -> Any {
+    internal func extractContent(medium: Content.Medium) throws -> Any {
         #if swift(>=6.0)
-        return try isolatedExtractContent(environmentObjects: environmentObjects)
+        return try isolatedExtractContent(medium: medium)
         #else
         return try MainActor.assumeIsolated {
-            try isolatedExtractContent(environmentObjects: environmentObjects)
+            try isolatedExtractContent(medium: medium)
         }
         #endif
     }
 
     @MainActor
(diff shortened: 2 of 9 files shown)


nalexn:
Thank you @nh7a

PR #429, opened 2026-08-20
Fix toolbar item inspection when wrapped by availability-gated ToolbarContent modifiers

### LINKED ISSUE #409: Toolbar `sharedBackgroundVisibility` modifier seems to break inspection
Hello there, I updated my code to add the iOS 26 `.sharedBackgroundVisibility(.hidden)` (tried with `.visible` value too) modifier, and the tests that were inspecting/interacting with the Toolbar are now failing. 

The modiier is used on `ToolbarItem` and `ToolbarItemGroup`, in both instances tests that were passing are failing.

I either get 
```
InvalidTransition { 
  phase: idle
  targetPhase: failed(deinit)
}"
```
or `search did not find a match` errors.

I have multiple tests with multiple ways of accessing the content inside the toolbar:
```Swift
view.find(viewWithAccessibilityLabel: "label")
try view.find(ViewType.Button.self, where: { try $0.accessibilityLabel().string() == "label" }).tap()
try view.find(CustomView.self).find(ViewType.ZStack.self).callOnTapGesture() // CustomView is a view inside ToolbarItem
```

I tried doing `view.find(ViewType.Toolbar.ItemGroup.self)` but the error also occurs.

I don't know f something needs to be updated in order for the inspection to work in this case or there's something else ? 
Thank you

tc-sgupta:
Fixes #409

## Problem

`if #available { ... } else { ... }` inside a `@ToolbarContentBuilder` compiles to `_ConditionalContent<TrueBranch, FalseBranch>`. When the true branch calls an availability-limited API (e.g. `sharedBackgroundVisibility`, introduced in iOS 26), the compiler further wraps it in `LimitedAvailabilityToolbarContent`, whose payload is boxed inside a single-element `TupleToolbarContent` and then a `ToolbarModifiedContent` pairing the original `ToolbarItem` with the applied modifier.

`toolbar().item(_:)`/`itemGroup(_:)` handed these wrapper values straight to `guardType`, which rejected them outright since none of them are literally `ToolbarItem`/`ToolbarItemGroup` — regardless of what the caller was searching for downstream. `find(ViewType.Button.self)`, `find(viewWithId:)`, and even `find(ViewType.Toolbar.ItemGroup.self)` directly all failed identically with `"Search did not find a match"`, matching every access pattern reported in #409.

## Fix

Adds `Inspector.resolveTransparentToolbarWrapper(_:)`, called from `item(_:)`/`itemGroup(_:)` before constructing the `InspectableView`, which peels through:
- `TupleToolbarContent` (single-element case)
- `_ConditionalContent`
- `LimitedAvailabilityToolbarContent`
- `ToolbarModifiedContent`

...to reach the real `ToolbarItem`/`ToolbarItemGroup`.

Resolution is best-effort: if an expected shape isn't found at any hop, the pre-hop value is returned as-is, so existing behavior for plain `ToolbarItem`/`ToolbarItemGroup` values is unaffected.

## Testing

Added 3 tests to `ToolbarTests.swift`:
- `.sharedBackgroundVisibility` applied directly to a `ToolbarItem`
- `.sharedBackgroundVisibility` applied directly to a `ToolbarItemGroup`
- The exact shape from #409 — the availability check lives in a separate `@ToolbarContentBuilder` helper extension applied inline, rather than at the `.toolbar { }` call site

Ran the full suite before and after this change: 92 pre-existing failures in both runs (unrelated to toolbars), confirming no regressions.

### DIFF
--- a/Sources/ViewInspector/SwiftUI/Toolbar.swift
+++ b/Sources/ViewInspector/SwiftUI/Toolbar.swift
@@ -103,12 +103,12 @@ public extension InspectableView where View == ViewType.Toolbar {
     }
 
     func item(_ index: Int = 0) throws -> InspectableView<ViewType.Toolbar.Item> {
-        let element = try self.element(index)
+        let element = try Inspector.resolveTransparentToolbarWrapper(self.element(index))
         return try .init(Content(element, medium: content.medium), parent: self, index: index)
     }
-    
+
     func itemGroup(_ index: Int = 0) throws -> InspectableView<ViewType.Toolbar.ItemGroup> {
-        let element = try self.element(index)
+        let element = try Inspector.resolveTransparentToolbarWrapper(self.element(index))
         return try .init(Content(element, medium: content.medium), parent: self, index: index)
     }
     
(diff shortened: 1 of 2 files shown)


nalexn:
Thanks for the PR!

PR #432, opened 2026-08-24
#425: Support TupleContent produced by ViewBuilder on iOS 27

### LINKED ISSUE #425: Xcode 27 Beta 5 test failures
I have a series of simple tests for some of my input fields. They pass with no issues on Xcode 26.6, but consistently fail in the Xcode 27 Beta versions.  I did try pinning directly to your "0.10.4" branch as I saw you had several Xcode 27 specific changes in it, but that did not change the results.

Here is what I am getting in the Xcode console:
_✘ Test labelText_isAlwaysPresent_whenEmpty() recorded an issue at FloatingLabelFieldTests.swift:14:6: Caught error: Search did not find a match. Possible blockers: TupleContent<Pack{ModifiedContent<ModifiedContent<ModifiedContent<Text, _OffsetEffect>, _AnimationModifier<Bool>>, AccessibilityAttachmentModifier>, ModifiedContent<ModifiedContent<ModifiedContent<ModifiedContent<ModifiedContent<ModifiedContent<ModifiedContent<ModifiedContent<TextField<Text>, FocusStateBindingModifier<Bool>>, _PaddingLayout>, _EnvironmentKeyWritingModifier<KeyboardType>>, _EnvironmentKeyTransformModifier<TextInputAutocapitalization>>, _EnvironmentKeyWritingModifier<Optional<Bool>>>, AccessibilityAttachmentModifier>, _ValueActionModifier2<Bool>>, _AppearanceActionModifier>}>_

Here is what Claude say's about the issue:
_What's happening: On the Xcode 27 beta SDK (you're on the xcode_27_beta_4 branch), SwiftUI emits parameter-pack-based tuple views for _multi-child ViewBuilder blocks, and ViewInspector 0.10.3 can't decompose them. The error message actually proves it: the "Possible blockers" dump is TupleContent<Pack{...Text..., ...TextField...}> — the "Brand" Text the test is looking for is right there inside the Pack, but ViewInspector can't step into it, so find(text:) comes up empty.__

nh7a:
Fixes #425

## Problem

With the iOS 27 SDK and a deployment target of iOS 27, inspection fails on any
container with two or more children:

    Search did not find a match. Possible blockers: TupleContent<Pack{Text, Text}>

## Cause

The iOS 27 SDK adds a `ViewBuilder.buildBlock` overload:

    @available(iOS 27.0, macOS 27.0, tvOS 27.0, watchOS 27.0, visionOS 27.0, *)
    public static func buildBlock<each Content>(_ content: repeat each Content)
        -> TupleContent<repeat each Content>

The pre-existing overload returning `TupleView` is `@_disfavoredOverload`, so in
an iOS 27 availability context `VStack { Text("A"); Text("B") }` is built as
`VStack<TupleContent<Pack{Text, Text}>>` instead of
`VStack<TupleView<(Text, Text)>>`. The children move from the `value` attribute
to `content`, and `Inspector.isTupleView` matched only `TupleView`, so
containers stopped decomposing. `ToolbarContentBuilder` changed the same way,
moving toolbar elements from `content|value|.N` to `content|content|.N`.

## Changes

* `ViewType.TupleContentView` — a distinct view type for `TupleContent`,
  inspected with `tupleContentView(_:)`, alongside `TupleView`/`tupleView(_:)`.
  Both share the extraction logic, differing only in the storage label.
* `Inspector.isViewTuple` for the checks that care about tuple-ness, with
  `isTupleView`/`isTupleContentView` for the specific types.
* The "please specify its index" blocker now names the matching call, so the
  suggestion is accurate on both SDKs.
* `Toolbar` picks the element path based on the container type.

## Testing

`TupleContentTests` mirror `TupleViewTests` and cover extraction, search,
modifier resetting and toolbar items. The fixtures are marked
`@available(iOS 27.0, ...)`, which raises the availability context so
`ViewBuilder` selects the `TupleContent` overload regardless of the deployment
target — they run on an iOS 27 simulator in a normal build, and skip below
iOS 27. They compile only under the Xcode 27 SDK (`#if compiler(>=6.4)`).

Existing tests that assert `.tupleView` in paths and error messages now derive
the expected call from the tuple type the builder actually produced, so they
pass under either deployment target.

Full suite, 0 failures:

| Toolchain | Destination |
| --- | --- |
| Xcode 26.6 | iOS 26.5, iOS 27.0, macOS |
| Xcode 27 | iOS 26.5, iOS 27.0, macOS |
| Xcode 27 | iOS 27.0 with `IPHONEOS_DEPLOYMENT_TARGET=27.0` |

Without the fix, that last configuration fails 278 tests; with the default
deployment target the SDK bump alone stays green, which is why the new tests
raise the availability context instead of relying on it.

### DIFF
--- a/Sources/ViewInspector/InspectableView.swift
+++ b/Sources/ViewInspector/InspectableView.swift
@@ -28,8 +28,9 @@ public struct InspectableView<View> where View: BaseViewType {
     
     private static func build(content: Content, parent: UnwrappedView?, call: String, index: Int?) throws -> Self {
         if !View.typePrefix.isEmpty,
-           Inspector.isTupleView(content.view),
-           View.self != ViewType.TupleView.self {
+           Inspector.isViewTuple(content.view),
+           View.self != ViewType.TupleView.self,
+           View.self != ViewType.TupleContentView.self {
             throw InspectionError.notSupported(
                 "Unable to extract \(View.typePrefix): please specify its index inside parent view")
         }
@@ -204,10 +205,11 @@ internal extension InspectableView where View: MultipleViewContent {
             throw InspectionError.viewIndexOutOfBounds(index: index, count: viewes.count)
         }
         let child = try viewes.element(at: index)
-        if !isTupleExtraction && Inspector.isTupleView(child.view) {
+        if !isTupleExtraction && Inspector.isViewTuple(child.view) {
+            let call = Inspector.isTupleContentView(child.view) ? "tupleContentView" : "tupleView"
             throw InspectionError.notSupported(
                 // swiftlint:disable:next line_length
-                "Please insert .tupleView(\(index)) after \(Inspector.typeName(type: View.self)) for inspecting its children at index \(index)")
+                "Please insert .\(call)(\(index)) after \(Inspector.typeName(type: View.self)) for inspecting its children at index \(index)")
         }
         return child
     }
--- a/Sources/ViewInspector/Inspector.swift
+++ b/Sources/ViewInspector/Inspector.swift
@@ -389,14 +389,26 @@ internal extension Inspector {
     #endif
     static func viewsInContainer(view: Any, medium: Content.Medium) throws -> LazyGroup<Content> {
         let unwrappedContainer = try Inspector.unwrap(content: Content(view, medium: medium.resettingViewModifiers()))
-        guard Inspector.isTupleView(unwrappedContainer.view) else {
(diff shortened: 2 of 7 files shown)


andy-vanwagoner-jn:
This fixes my test suite in Xcode & iOS 27.

andy-vanwagoner-jn (approved):

nalexn:
Thanks for the PR!

PR #433, opened 2026-08-27
#363: Fix crash when search traverses a Charts `Chart` view

### LINKED ISSUE #363: Find API is crashing when iterating over `Chart` view
Hi Team,

After taking release of 0.10.0 our find view api is crashing every time. I already assign required environment object etc but still it's causing the issue.

Example: let view = try? subject.inspect().find(MyView.self)

nh7a:
Fixes #363

## What

Registers `Chart` as a known leaf view type so the view search identifies it and stops there, instead of unwrapping its body.

## Why

`Chart`'s body is built from a `GeometryReader` whose content closure requires a real layout context. ViewInspector feeds every `GeometryReader` a zero-filled `GeometryProxy`, and Charts then traps on an implicitly unwrapped nil inside its private `RenderBasedChartView`, taking down the whole test process.

Any `find` that has to walk *past* a chart crashes; a `find` that matches before reaching the chart happens to survive, which is why the failure looks intermittent:

```swift
struct ChartView: View {
    var body: some View {
        VStack {
            Chart { BarMark(x: .value("a", 1), y: .value("b", 2)) }.id("chart")
        }
    }
}

let sut = try ChartView().inspect()
try? sut.find(viewWithId: "chart")   // fine — matched before descending
try? sut.find(viewWithId: "text")    // crashed — walked into the chart's body
```

Backtrace of the crash: `Charts` frame called directly from `GeometryReader.view()` (`GeometryReader.swift:51`), where the fabricated `GeometryProxy` is passed to the content builder.

## Notes for the reviewer

- Nothing is lost by not descending: a chart's contents are `ChartContent`, not `View`, so they were never inspectable.
- No `import Charts` is needed — the type is matched by its namespaced name (`Charts.Chart`), so the code compiles and behaves the same on platforms without the framework.
- Adds `chart()` / `chart(_ index:)` extraction, matching the `Map` / `VideoPlayer` pattern for opaque framework views.
- `readiness.md` updated.
- Verified on Xcode 27.0 (27A266a): the crash still reproduces without this change, and the full suite passes with it — 1605 tests, 7 skipped, 0 failures.
- The tests deliberately use charts containing a `BarMark`. An empty `Chart { }` segfaults inside the Charts framework at construction time on this SDK, with no ViewInspector involved:
  ```swift
  let c = Chart { }   // EXC_BAD_ACCESS
  ```
  That looks like a separate Apple bug and is out of scope here.

### DIFF
--- a/Sources/ViewInspector/SwiftUI/Chart.swift
+++ b/Sources/ViewInspector/SwiftUI/Chart.swift
@@ -1,0 +1,38 @@
+import SwiftUI
+
+@available(iOS 13.0, macOS 10.15, tvOS 13.0, *)
+public extension ViewType {
+
+    /// `Chart` is an opaque view: its `body` is built from `GeometryReader` and other
+    /// internals that require a real layout context, so ViewInspector does not descend
+    /// into it. The chart's content is `ChartContent`, not `View`, and is not inspectable.
+    struct Chart: KnownViewType {
+        public static let typePrefix: String = "Chart"
+        public static var namespacedPrefixes: [String] {
+            return ["Charts." + typePrefix]
+        }
+        public static func inspectionCall(typeName: String) -> String {
+            return "chart(\(ViewType.indexPlaceholder))"
+        }
+    }
+}
+
+// MARK: - Extraction from SingleViewContent parent
+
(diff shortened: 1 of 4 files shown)


nalexn:
Thanks for the PR!

PR #435, opened 2026-09-10
Fix accessibilityTraits() throwing for an empty trait set

nh7a:
## Problem

`accessibilityTraits()` throws `modifierNotFound` when a view's only trait modifier resolves to an empty set:

```swift
Text("x").accessibilityAddTraits([])                          // throws
Text("x").accessibilityAddTraits(cond ? [.isSelected] : [])   // throws on the false arm
```

```
ViewInspector.InspectionError.modifierNotFound:
  "Text does not have 'accessibilityAddTraits' modifier"
```

The second form is the one that bites in practice — a test covering both arms of a conditional trait passes for one branch and throws for the other.

## Cause

A SwiftUI **runtime** change, not a long-standing bug.

On the iOS 26 runtime, `AccessibilityProperties` is a struct with a named `traits` field that is populated even for an empty set (`value: 0, mask: 0`), so the existing lookup finds it and returns `AccessibilityTraits()`:

```
properties: AccessibilityProperties
  label: Optional<AccessibilityLabelStorage>
  traits: Optional<AccessibilityNullableOptionSet<AccessibilityTraitSet>>
    some: … value.rawValue = 0, mask.rawValue = 0
```

On the iOS 27 runtime the properties moved to a type-keyed collection, and an empty trait set contributes **no entry at all** — there is nothing left for the reflection walk to find:

```
properties: AccessibilityProperties
  storage: Array<Entry> = []            // .accessibilityAddTraits([])
  storage: Array<Entry> = [Entry(key: TraitsKey, value: …)]   // .accessibilityAddTraits(.isButton)
```

It reproduces under both Xcode 26.6 and Xcode 27 whenever the iOS 27 runtime is used, which confirms the trigger is the runtime rather than the SDK.

## Fix

The `AccessibilityAttachmentModifier` itself is still emitted, so the empty set is recoverable. Every other accessibility modifier stores a value under its own key — `LabelKey`, `VisibilityKey`, `IdentifierKey`, `ValueStorageKey`, `HintsKey`, `InputLabelsKey`, `SortPriorityKey`, `ActionsKey` — including the no-op-looking ones (`accessibilityHidden(false)`, `accessibilityLabel("")`, `accessibilityInputLabels([])`, `accessibility(sortPriority: 0)`). That makes an attachment carrying *no* accessibility properties whatsoever the signature of a trait modifier that contributed no bits.

So when no trait value is found, `accessibilityTraits()` now returns `AccessibilityTraits()` if some attachment modifier has an empty property collection, and throws otherwise. The check keys off the collection shape rather than an OS version, so it cannot fire on the iOS 26 layout, where the property bag is a struct.

The throw is preserved where it should be:

- a view with no accessibility modifier at all (`padding()` emits no attachment modifier),
- a view whose only accessibility modifier stores some other property — `testMissingAccessibilityTraits` uses `accessibility(label:)` and still throws on every runtime.

## Known limitation

On the iOS 27 runtime, an empty trait modifier combined with another accessibility modifier is coalesced into a single attachment carrying only the other modifier's key:

```swift
EmptyView().accessibilityAddTraits([]).accessibility(label: Text("abc"))
// → one attachment, storage: [Entry(key: LabelKey, …)]
```

That is byte-identical to `accessibility(label:)` applied alone, which must keep throwing. Nothing distinguishes the two, so this case still throws rather than weakening the throw for a view that genuinely has no trait modifier.

## Tests

Two tests added, both written to fail before the fix:

- `testAccessibilityEmptyTraitsInspection` — empty add and empty remove, both spellings (`accessibility(addTraits:)` / `accessibilityAddTraits`), on `EmptyView` and `Text`.
- `testAccessibilityEmptyAndNonEmptyTraitsInspection` — empty add + non-empty remove, non-empty add + empty remove, and both empty.

Pre-fix, with the tests in place:

| | Xcode 27 / iOS 27.0 | Xcode 26.6 / iOS 26.5 |
|---|---|---|
| both tests | fail | pass |

Post-fix, full suite (1563 tests, 15 skipped, 0 failures in every cell):

| Toolchain | iOS 26.5 | iOS 27.0 |
|---|---|---|
| Xcode 27.0 (27A266a) | pass | pass |
| Xcode 26.6 (17F113) | pass | pass |

### DIFF
--- a/Sources/ViewInspector/Modifiers/AccessibilityTraitsModifiers.swift
+++ b/Sources/ViewInspector/Modifiers/AccessibilityTraitsModifiers.swift
@@ -14,9 +14,14 @@ public extension InspectableView {
         let call = "accessibilityAddTraits"
         let traitSets = accessibilityTraitSets()
         guard !traitSets.isEmpty else {
-            throw InspectionError
-                .modifierNotFound(parent: Inspector.typeName(value: content.view),
-                                  modifier: call, index: 0)
+            /* A trait modifier that resolves to an empty set may leave no trait value
+               behind at all, in which case the empty set is still the correct answer. */
+            guard hasPropertylessAccessibilityAttachment() else {
+                throw InspectionError
+                    .modifierNotFound(parent: Inspector.typeName(value: content.view),
+                                      modifier: call, index: 0)
+            }
+            return AccessibilityTraits()
         }
         let rawValue = traitSets.reduce(UInt64(0)) { result, traitSet in
             (result & ~traitSet.mask) | (traitSet.value & traitSet.mask)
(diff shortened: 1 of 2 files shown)


nalexn:
Thanks for the PR!

